PR #45, opened 2023-09-12
CODEOWNERS, Migrations and Test Corrections

ralikio:
**Description**

Changes proposed in this pull request:

- Adding Framefrogs as owners of subscriptioncleanup job,
- added missing migrations from old repository,
- changed migrations path in e2e tests.

**Related issue(s)**

PR #47, opened 2023-09-12
Mov kyma env cleanup job2

ukff:


ralikio:
Migrations scripts done in https://github.com/kyma-project/kyma-environment-broker/pull/45

PR #65, opened 2023-09-22
Fix AvS problems with optimistic locking

piotrmiskiewicz:
**Description**

- AvS makes a retry in case of optimistic locking error (conflict)
- stagedmanager runs the step with the newest operation

The staged manager implementation was working fine except a case, when step saves the operation, but then returns a retry. Probably AvS is the only one such case.

**Related issue(s)**

### DIFF
--- a/internal/process/deprovisioning/avs_evaluations.go
+++ b/internal/process/deprovisioning/avs_evaluations.go
@@ -1,60 +1,60 @@
 package deprovisioning
 
 import (
-    "time"
+	"time"
 
-    "github.com/kyma-project/kyma-environment-broker/internal/broker"
+	"github.com/kyma-project/kyma-environment-broker/internal/broker"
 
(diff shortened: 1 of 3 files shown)


internal/process/deprovisioning/avs_evaluations.go
```diff
@@ -15,16 +15,15 @@ import (
 
 type AvsEvaluationRemovalStep struct {
 	delegator             *avs.Delegator
-	operationsStorage     storage.Operations
 	externalEvalAssistant avs.EvalAssistant
 	internalEvalAssistant avs.EvalAssistant
 	deProvisioningManager *process.OperationManager
+	operationManager      *process.OperationManager
```
szwedm, line 21: I couldn't find usages of `operationManager`, looks redundant.

szwedm (approved):
/lgtm

PR #90, opened 2023-10-06
Make region always mandatory in the KEB schema

MarekMichali:
**Description**

**Description**

- Make schema creation more generic
- Generate schema with required `region` 
- Add a feature flag
- Adjust tests

**Related issue(s)**

See also https://github.com/kyma-project/control-plane/issues/3006


internal/appinfo/runtime_info.go
```diff
@@ -30,20 +30,22 @@ type (
...
-		lastOperationFinder:     lastOpFinder,
-		respWriter:              respWriter,
-		plansConfig:             plansConfig,
-		defaultSubaccountRegion: region,
+		instanceFinder:            instanceFinder,
+		lastOperationFinder:       lastOpFinder,
+		respWriter:                respWriter,
+		plansConfig:               plansConfig,
+		defaultSubaccountRegion:   region,
+		regionParameterIsRequired: regionParameterIsRequired,
```
piotrmiskiewicz, line 48: you don't need this bool flag

internal/appinfo/runtime_info.go
```diff
@@ -157,7 +159,7 @@ func (h *RuntimeInfoHandler) planNameOrDefault(inst internal.InstanceWithOperati
 	if inst.ServicePlanName != "" {
 		return inst.ServicePlanName
 	}
-	return broker.Plans(h.plansConfig, "", false, false)[inst.ServicePlanID].Name
+	return broker.Plans(h.plansConfig, "", false, false, h.regionParameterIsRequired)[inst.ServicePlanID].Name //tu
```
piotrmiskiewicz, line 162: you are taking only a name here, the flag which indicates if parameter is required is not necessary here

internal/appinfo/runtime_info.go
MarekMichali: I will remove it.

internal/appinfo/runtime_info.go
MarekMichali: You're right, I didn't see that we only take the name parameter.

internal/appinfo/runtime_info.go
```diff
@@ -37,7 +37,7 @@ type RuntimeInfoHandler struct {
 	defaultSubaccountRegion string
 }
 
-func NewRuntimeInfoHandler(instanceFinder InstanceFinder, lastOpFinder LastOperationFinder, plansConfig broker.PlansConfig, region string, respWriter ResponseWriter) *RuntimeInfoHandler {
+func NewRuntimeInfoHandler(instanceFinder InstanceFinder, lastOpFinder LastOperationFinder, plansConfig broker.PlansConfig, region string, respWriter ResponseWriter, regionParameterIsRequired bool) *RuntimeInfoHandler {
```
piotrmiskiewicz, line 40: regionParameterIsRequired is not used

MarekMichali: Removed.

piotrmiskiewicz (approved):

PR #116, opened 2023-10-20
Fix region for kyma CR

vvxxvvxx:
**Description**

- Fix the issue that the region value is missing when it's assigned to Kyma CR as the region is from `operation.Region` and this `operation.Region` currently is from [operation.RuntimeOperation.Runtime.Region](https://github.com/kyma-project/kyma-environment-broker/blob/e8cdf9dc1e84c723cb727daa7d56693beccc6f02/common/orchestration/ext.go#L21). it's missing during provisioning.
The region is actually assigned to [requestInput.ClusterConfig.GardenerConfig.Region](https://github.com/kyma-project/kyma-environment-broker/blob/e8cdf9dc1e84c723cb727daa7d56693beccc6f02/internal/process/provisioning/create_runtime_without_kyma_step.go#L63C3-L63C51) already in the step `Create_Runtime_Without_Kyma` and this `requestInput.ClusterConfig.GardenerConfig.Region` is assigned from [r.provisioningParameters.Parameters](https://github.com/kyma-project/kyma-environment-broker/blob/e8cdf9dc1e84c723cb727daa7d56693beccc6f02/internal/process/input/input.go#L458).
The `r.provisioningParameters.Parameters` is set in [operation.InputCreator.SetProvisioningParameters](https://github.com/kyma-project/kyma-environment-broker/blob/e8cdf9dc1e84c723cb727daa7d56693beccc6f02/internal/process/provisioning/create_runtime_without_kyma_step.go#L143) from `operation.ProvisioningParameters`. So here use the region from `operation.ProvisioningParameters.Parameters.Region` directly.

**Related issue(s)**
related to [the issue](https://github.tools.sap/kyma/backlog/issues/4575)

### DIFF
--- a/internal/process/steps/lifecycle_manager.go
+++ b/internal/process/steps/lifecycle_manager.go
@@ -23,7 +23,11 @@ func ApplyLabelsAndAnnotationsForLM(object client.Object, operation internal.Ope
 	l["kyma-project.io/global-account-id"] = operation.GlobalAccountID
 	l["kyma-project.io/subaccount-id"] = operation.SubAccountID
 	l["kyma-project.io/shoot-name"] = operation.ShootName
-	l["kyma-project.io/region"] = *operation.ProvisioningParameters.Parameters.Region
+	if operation.ProvisioningParameters.Parameters.Region != nil {
+		l["kyma-project.io/region"] = *operation.ProvisioningParameters.Parameters.Region
+	} else {
+		l["kyma-project.io/region"] = ""
+	}
 	l["operator.kyma-project.io/kyma-name"] = KymaName(operation)
 	l["operator.kyma-project.io/managed-by"] = "lifecycle-manager"
 	if isKymaResourceInternal(operation) {
--- a/internal/process/steps/lifecycle_manager.go
+++ b/internal/process/steps/lifecycle_manager.go
@@ -23,7 +23,7 @@ func ApplyLabelsAndAnnotationsForLM(object client.Object, operation internal.Ope
 	l["kyma-project.io/global-account-id"] = operation.GlobalAccountID
 	l["kyma-project.io/subaccount-id"] = operation.SubAccountID
 	l["kyma-project.io/shoot-name"] = operation.ShootName
-	l["kyma-project.io/region"] = operation.Region
+	l["kyma-project.io/region"] = *operation.ProvisioningParameters.Parameters.Region
 	l["operator.kyma-project.io/kyma-name"] = KymaName(operation)
 	l["operator.kyma-project.io/managed-by"] = "lifecycle-manager"
 	if isKymaResourceInternal(operation) {


piotrmiskiewicz:
operation.RuntimeOperation.Runtime.Region is not set in the provisioning step. On the other hand, provisioning parameters could not contain the region - in such case the default one is taken. In my opinion it does not make sense. 
@jaroslaw-pieszka  is taking care of this issue

The first solution is: The create_runtime_without_kyma step knows the region (which is stored in the instance table). We can update the operation in that step - save the region also in the operation (but in the InstanceDetails). InstanceDetails are always copied to the new operation, so after provisioning, every operation will have it.

Second solution: If we want to have region for existing SKRs, we should copy the region from the Instance in constructors (NewUpdateOperation etc.). The constructors have instance as argument so we can copy the region. After such change every new operation will contain the region and everytime when the apply kyma step is executed, the label will be applied

vvxxvvxx:
Hi @piotrmiskiewicz , the create_runtime_without_kyma step knows the region (which is [stored in the instance table](https://github.com/kyma-project/kyma-environment-broker/blob/e8cdf9dc1e84c723cb727daa7d56693beccc6f02/internal/process/provisioning/create_runtime_without_kyma_step.go#L104)). this `requestInput.ClusterConfig.GardenerConfig.Region` is copied from [r.provisioningParameters.Parameters](https://github.com/kyma-project/kyma-environment-broker/blob/e8cdf9dc1e84c723cb727daa7d56693beccc6f02/internal/process/input/input.go#L458). and this `r.provisioningParameters.Parameters` is copied from [operation.ProvisioningParameters](https://github.com/kyma-project/kyma-environment-broker/blob/e8cdf9dc1e84c723cb727daa7d56693beccc6f02/internal/process/provisioning/create_runtime_without_kyma_step.go#L143). So is `operation.ProvisioningParameters` also stored in storage?

ebensom:
@vvxxvvxx @vvxxvvxx why not always take it from [operation.LastRuntimeState.ClusterConfig.Region](https://github.com/kyma-project/kyma-environment-broker/blob/main/internal/model.go#L415C31-L415C31) as `Create_Runtime` step persist it there?

vvxxvvxx:
> @vvxxvvxx @piotrmiskiewicz why not always take it from [operation.LastRuntimeState.ClusterConfig.Region](https://github.com/kyma-project/kyma-environment-broker/blob/main/internal/model.go#L415C31-L415C31) as `Create_Runtime` step persist it there?

When creating a new type operation, seems it will always get [instanceStorage.GetByID(instanceID)](https://github.com/kyma-project/kyma-environment-broker/blob/c329e328d31d724ff2de7ce8932d8dca2d931fac/internal/broker/instance_update.go#L90C21-L90C56) first for the info for Operation. we could see this for [deprovisioning operation](https://github.com/kyma-project/kyma-environment-broker/blob/c329e328d31d724ff2de7ce8932d8dca2d931fac/internal/broker/instance_deprovision.go#L45), [update operation](https://github.com/kyma-project/kyma-environment-broker/blob/c329e328d31d724ff2de7ce8932d8dca2d931fac/internal/broker/instance_update.go#L90C21-L90C56) and [orchestration operation](https://github.com/kyma-project/kyma-environment-broker/blob/c387bfcad4d5b90610da79ee21e657aea6e80dc0/internal/orchestration/manager/manager.go#L213), ect..
but this [internal.Instance struct](https://github.com/kyma-project/kyma-environment-broker/blob/c387bfcad4d5b90610da79ee21e657aea6e80dc0/internal/model.go#L124) doesn't contain `RuntimeState`. so when creating a new type operation, this `RuntimeState` is not assigned to Operation in the first place. But the `internal.Instance` contains [ProvisioningParameters](https://github.com/kyma-project/kyma-environment-broker/blob/c387bfcad4d5b90610da79ee21e657aea6e80dc0/internal/model.go#L136) and seems it's always assigned to Operation.

and during the provisioning, this persisted [requestInput.ClusterConfig.GardenerConfig.Region](https://github.com/kyma-project/kyma-environment-broker/blob/e8cdf9dc1e84c723cb727daa7d56693beccc6f02/internal/process/provisioning/create_runtime_without_kyma_step.go#L104) is also copied from [provisioningParameters.Parameters](https://github.com/kyma-project/kyma-environment-broker/blob/e8cdf9dc1e84c723cb727daa7d56693beccc6f02/internal/process/input/input.go#L458) ( [operation.ProvisioningParameters](https://github.com/kyma-project/kyma-environment-broker/blob/e8cdf9dc1e84c723cb727daa7d56693beccc6f02/internal/process/provisioning/create_runtime_without_kyma_step.go#L143)).
so i think we could use `*operation.ProvisioningParameters.Parameters.Region` here directly in this PR..

jaroslaw-pieszka:
Fixed in https://github.com/kyma-project/kyma-environment-broker/pull/123

PR #118, opened 2023-10-20
Fix runtime dto Status.State for deprovision type

vvxxvvxx:
**Description**

- Adjust the Status.State for deprovision type in Runtime DTO as the deprovisioning operation state can be `domain.Succeeded` when it still contains non empty executedButNotCompletedSteps. so as long as the latest operation is in type `Deprovision` and state is `Succeeded`, turn its runitme Status.State to `StateDeprovisioning`.
 it's now causing [the mop-agent issue](https://github.tools.sap/kyma/backlog/issues/4564) (failed [in the check here](https://github.tools.sap/kyma/mop-agent/blob/6b86990ec7f5e1ccb625d071af9ddfeb5b445917/internal/runtime/manager.go#L205) and mop-agent still trying to send request to AVS for the deprovisioning runtime causing `400: bad request`).

**Related issue(s)**

### DIFF
--- a/internal/runtime/converter.go
+++ b/internal/runtime/converter.go
@@ -198,7 +198,7 @@ func (c *converter) adjustRuntimeState(dto *pkg.RuntimeDTO) {
 		case pkg.Suspension:
 			dto.Status.State = pkg.StateSuspended
 		case pkg.Deprovision:
-			if len(lastOp.ExcutedButNotCompleted) == 0 {
+			if len(lastOp.ExecutedButNotCompletedSteps) == 0 {
 				dto.Status.State = pkg.StateDeprovisioned
 			} else {
 				dto.Status.State = pkg.StateDeprovisionIncomplete
--- a/internal/runtime/converter.go
+++ b/internal/runtime/converter.go
@@ -194,8 +194,11 @@ func (c *converter) adjustRuntimeState(dto *pkg.RuntimeDTO) {
 	switch lastOp.State {
 	case string(domain.Succeeded):
 		dto.Status.State = pkg.StateSucceeded
-		if lastOp.Type == pkg.Suspension {
+		switch lastOp.Type {
+		case pkg.Suspension:
 			dto.Status.State = pkg.StateSuspended
+		case pkg.Deprovision:
+			dto.Status.State = pkg.StateDeprovisioning
 		}
 	case string(domain.Failed):
 		dto.Status.State = pkg.StateFailed
--- a/internal/runtime/converter.go
+++ b/internal/runtime/converter.go
@@ -198,7 +198,11 @@ func (c *converter) adjustRuntimeState(dto *pkg.RuntimeDTO) {
 		case pkg.Suspension:
 			dto.Status.State = pkg.StateSuspended
 		case pkg.Deprovision:
-			dto.Status.State = pkg.StateDeprovisioning
+			if len(lastOp.ExcutedButNotCompleted) == 0 {
+				dto.Status.State = pkg.StateDeprovisioned
+			} else {
+				dto.Status.State = pkg.StateDeprovisionIncomplete
+			}
 		}
 	case string(domain.Failed):
 		dto.Status.State = pkg.StateFailed


piotrmiskiewicz:
Hi, the endpoint is designed for KCP CLI and how your changes affects. Currently, the SKR which is not completely deprovisioned (some step needs a retry) looks like:
```
./kcp-darwin rt -S deprovisionIncomplete
GLOBALACCOUNT ID                       SUBACCOUNT ID          SHOOT       REGION         PLAN   CREATED AT            STATE       EU ACCESS
e449f875-b5b2-4485-b7c0-98725c0571bf   prow-keb-integration   c-46fb007   eu-central-1   aws    2023/10/20 01:07:45   succeeded   No
e449f875-b5b2-4485-b7c0-98725c0571bf   prow-keb-integration   ba915ad     eu-central-1   aws    2023/10/20 01:56:26   succeeded   No
```

after this PR is merged, it will look like:
```
GLOBALACCOUNT ID                       SUBACCOUNT ID          SHOOT       REGION         PLAN   CREATED AT            STATE       EU ACCESS
e449f875-b5b2-4485-b7c0-98725c0571bf   prow-keb-integration   c-46fb007   eu-central-1   aws    2023/10/20 01:07:45   deprovisioning   No
e449f875-b5b2-4485-b7c0-98725c0571bf   prow-keb-integration   ba915ad     eu-central-1   aws    2023/10/20 01:56:26   deprovisioning   No
```

internal/runtime/converter.go
```diff
@@ -194,8 +194,11 @@ func (c *converter) adjustRuntimeState(dto *pkg.RuntimeDTO) {
 	switch lastOp.State {
 	case string(domain.Succeeded):
 		dto.Status.State = pkg.StateSucceeded
-		if lastOp.Type == pkg.Suspension {
+		switch lastOp.Type {
+		case pkg.Suspension:
 			dto.Status.State = pkg.StateSuspended
+		case pkg.Deprovision:
+			dto.Status.State = pkg.StateDeprovisioning
```
ebensom, line 201: In this case the runtime is not in deprovisioning lifecycle state.

IMO the logic should be here:
- `if lastOp.type == Suspension -> StateSuspended`
- if lastOp.type == Deprovision && !dto.Status.DeletedAt.IsZero() -> [`StateDeprovisionIncomplete`](https://github.com/kyma-project/kyma-environment-broker/blob/main/common/runtime/model.go#L26)
- if lastOp.type == Deprovision && dto.Status.DeletedAt.IsZero() -> StateDeprovisioned

ebensom (changes requested):

internal/runtime/converter.go
ebensom: @vvxxvvxx this also means the mop-agent code must be updated to be able to cope with the [StateDeprovisionIncomplete](https://github.com/kyma-project/kyma-environment-broker/blob/main/common/runtime/model.go#L26) state. In this lifecycle state, the runtime is of course excluded from AvS monitoring.

piotrmiskiewicz:
@vvxxvvxx please contact with me

ebensom:
@piotrmiskiewicz I think this PR makes sense, because using Succeeded state for deprovisioned runtimes is not good. I would propose to use `deprovisionincomplete` state if there are any unfinished stages, otherwise use `deprovisioned`.

vvxxvvxx:
thanks for the comments. what about changing to the logic as below?
```
	case string(domain.Succeeded):
		dto.Status.State = pkg.StateSucceeded
		switch lastOp.Type {
		case pkg.Suspension:
			dto.Status.State = pkg.StateSuspended
		case pkg.Deprovision
		    if len(lastOp.ExcutedButNotCompleted) ==0 {
			    dto.Status.State = pkg.StateDeprovisioned
			} else {
			    dto.Status.State = pkg.StateDeprovisionIncomplete
			}		
		}
```

ebensom:
@vvxxvvxx this looks good to me!

vvxxvvxx:
@piotrmiskiewicz suggested in mop-agent, if we call KEB without any options, we won't get the deprovision runtimes. 
Now in mop-agent, we already set [the KEB call without any options](https://github.tools.sap/kyma/mop-agent/blob/6b86990ec7f5e1ccb625d071af9ddfeb5b445917/internal/api/runtime.go#L54).
Further checked the codes, in KEB, when getting `/runtimes` req without any state option, it will [exclude the deprovisioned instance](https://github.com/kyma-project/kyma-environment-broker/blob/c387bfcad4d5b90610da79ee21e657aea6e80dc0/internal/runtime/handler.go#L421C49-L421C73). and for the filter state `dbmodel.InstanceNotDeprovisioned`, it's handled [here](https://github.com/kyma-project/kyma-environment-broker/blob/c387bfcad4d5b90610da79ee21e657aea6e80dc0/internal/storage/postsql/read.go#L776) . But now mop-agent can still receive the deprovisioning runtimes in state `domain.Succeeded` but with non-empty `ExcutedButNotCompleted` steps. I think this is caused by that when the deprovisioning operation is retried, in the operation, its state becomes [domain.InProgress](https://github.com/kyma-project/kyma-environment-broker/blob/c329e328d31d724ff2de7ce8932d8dca2d931fac/internal/process/deprovisioning/init.go#L72) again. so in the retry, the deprovisioning operation becomes `type = internal.OperationTypeDeprovision` and `state= domain.InProgress` and it can't be excluded by the KEB call without any options.. then mop-agent can still receive these retrying deprovisioning operations which cause the mop-agent [AVS status: 400 Bad Request #4564](https://github.tools.sap/kyma/backlog/issues/4564) issue.
So the problem is how we could exclude these runtimes having retrying `InProgress` deprovisioning operation?

vvxxvvxx:
the deprovisioning operations can't be excluded with `AllOperation` option. after discussion with @piotrmiskiewicz , we could analyze the runtime DTO returned and exclude the deprovision type runtime in mop-agent

vvxxvvxx:
the test pre-kyma-environment-broker-presubmit-scanner failed in the gitleak check for [the line](https://github.com/kyma-project/kyma-environment-broker/blob/ebbc2f7869eb72ae4da87ceda9b17c92e758ed05/resources/keb/values.yaml#L132C3-L132C51). checking with @ralikio

ralikio (approved):

PR #129, opened 2023-10-30
Add a flag to expose schema with region required

MarekMichali:
**Description**

The flag allows to expose a new schema for KEB that makes the region parameter mandatory, but KEB will still accept requests without the region parameter.

- Add a flag to expose schema with the region required parameter
- Add unit tests

**Related issue(s)**

See also https://github.com/kyma-project/control-plane/issues/3006

### DIFF
--- a/internal/broker/services_test.go
+++ b/internal/broker/services_test.go
@@ -116,7 +116,7 @@ func TestServices_Services(t *testing.T) {
 		assertPlansContainPropertyInSchemas(t, services[0], "administrators")
 	})
 
-	t.Run("should containt property required with values name and region", func(t *testing.T) {
+	t.Run("should containt the property 'required' with values [name region] when ExposeSchemaWithRegionRequired is true and RegionParameterIsRequired is false", func(t *testing.T) {
 		// given
 		var (
 			name       = "testServiceName"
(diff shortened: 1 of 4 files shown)


internal/broker/services_test.go
```diff
@@ -115,6 +115,166 @@ func TestServices_Services(t *testing.T) {
 		assertPlansContainPropertyInSchemas(t, services[0], "oidc")
 		assertPlansContainPropertyInSchemas(t, services[0], "administrators")
 	})
+
+	t.Run("should containt the property 'required' with values [name region] when ExposeSchemaWithRegionRequired is true and RegionParameterIsRequired is false", func(t *testing.T) {
```
MarekMichali, line 119: Any unit tests added in this PR may be removed after this feature goes live on all environments.

jaroslaw-pieszka, line 119:
```suggestion
	t.Run("should contain the property 'required' with values [name region] when ExposeSchemaWithRegionRequired is true and RegionParameterIsRequired is false", func(t *testing.T) {
```

internal/broker/services_test.go
```diff
@@ -115,6 +115,166 @@ func TestServices_Services(t *testing.T) {
...
+		assert.Equal(t, name, services[0].Metadata.DisplayName)
+		assert.Equal(t, supportURL, services[0].Metadata.SupportUrl)
+
+		for _, plan := range services[0].Plans {
+			assertPlanContainsPropertyValuesInCreateSchema(t, plan, "required", []string{"name", "region"})
+		}
+
+	})
+
+	t.Run("should containt the property 'required' with values [name region] when ExposeSchemaWithRegionRequired is true and RegionParameterIsRequired is true", func(t *testing.T) {
```
jaroslaw-pieszka, line 159:
```suggestion
	t.Run("should contain the property 'required' with values [name region] when ExposeSchemaWithRegionRequired is true and RegionParameterIsRequired is true", func(t *testing.T) {
```

internal/broker/services_test.go
```diff
@@ -115,6 +115,166 @@ func TestServices_Services(t *testing.T) {
...
+		assert.Equal(t, name, services[0].Metadata.DisplayName)
+		assert.Equal(t, supportURL, services[0].Metadata.SupportUrl)
+
+		for _, plan := range services[0].Plans {
+			assertPlanContainsPropertyValuesInCreateSchema(t, plan, "required", []string{"name", "region"})
+		}
+
+	})
+
+	t.Run("should containt the property 'required' with values [name region] when ExposeSchemaWithRegionRequired is false and RegionParameterIsRequired is true", func(t *testing.T) {
```
jaroslaw-pieszka, line 199:
```suggestion
	t.Run("should contain the property 'required' with values [name region] when ExposeSchemaWithRegionRequired is false and RegionParameterIsRequired is true", func(t *testing.T) {
```

internal/broker/services_test.go
```diff
@@ -115,6 +115,166 @@ func TestServices_Services(t *testing.T) {
...
+		assert.Equal(t, name, services[0].Metadata.DisplayName)
+		assert.Equal(t, supportURL, services[0].Metadata.SupportUrl)
+
+		for _, plan := range services[0].Plans {
+			assertPlanContainsPropertyValuesInCreateSchema(t, plan, "required", []string{"name", "region"})
+		}
+
+	})
+
+	t.Run("should containt the property 'required' with values [name] when ExposeSchemaWithRegionRequired is false and RegionParameterIsRequired is false", func(t *testing.T) {
```
jaroslaw-pieszka, line 239:
```suggestion
	t.Run("should contain the property 'required' with values [name] when ExposeSchemaWithRegionRequired is false and RegionParameterIsRequired is false", func(t *testing.T) {
```

internal/broker/services_test.go
```diff
@@ -115,6 +115,166 @@ func TestServices_Services(t *testing.T) {
...
+		// then
+		require.NoError(t, err)
+		assert.Len(t, services, 1)
+		assert.Len(t, services[0].Plans, 5)
+
+		assert.Equal(t, name, services[0].Metadata.DisplayName)
+		assert.Equal(t, supportURL, services[0].Metadata.SupportUrl)
+
+		for _, plan := range services[0].Plans {
+			assertPlanContainsPropertyValuesInCreateSchema(t, plan, "required", []string{"name"})
```
jaroslaw-pieszka, line 274: Do you check if there is only `name` or if the list of required contains `name`?

MarekMichali: I check if there is only `name`

MarekMichali: For example, if we pass []string{"name","region"}, we will get that the schema doesn't contain the `region`. For example, if we pass an empty slice, we will get that the schema has more values than we passed.

internal/broker/services_test.go
```diff
@@ -139,3 +299,25 @@ func assertPlanContainsPropertyInUpdateSchema(t *testing.T, plan domain.ServiceP
 		t.Errorf("plan %s does not contain %s property in Update schema", plan.Name, property)
 	}
 }
+
+func assertPlanContainsPropertyValuesInCreateSchema(t *testing.T, plan domain.ServicePlan, property string, wantedPropertyValues []string) {
```
jaroslaw-pieszka, line 303: Could we use some asserts from the library?
We just compare string with array of strings.

jaroslaw-pieszka (approved):

jaroslaw-pieszka (approved):

PR #137, opened 2023-11-02
fixing provider tag value

i332322:
fix provider tag value.

### DIFF
--- a/internal/avs/external_eval_assistant.go
+++ b/internal/avs/external_eval_assistant.go
@@ -68,7 +68,7 @@ func (eea *ExternalEvalAssistant) ProvideTags(operation internal.Operation) []*T
 			TagClassId: eea.avsConfig.LandscapeTagClassId,
 		},
 		{
-			Content:    providerCodeByPlan(operation.ProvisioningParameters.PlanID),
+			Content:    providerCodeByPlan(operation.ProvisioningParameters.PlanID, true),
 			TagClassId: eea.avsConfig.ProviderTagClassId,
 		},
 		{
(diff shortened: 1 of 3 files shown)


ebensom (approved):

internal/avs/external_eval_assistant.go
```diff
@@ -68,7 +68,7 @@ func (eea *ExternalEvalAssistant) ProvideTags(operation internal.Operation) []*T
 			TagClassId: eea.avsConfig.LandscapeTagClassId,
 		},
 		{
-			Content:    string(operation.ProvisioningParameters.PlatformProvider),
+			Content:    providerCodeByPlan(operation.ProvisioningParameters.PlanID),
```
ebensom, line 71: The `providerCodeByPlan` returns "AZR" for azure plan, however our provider tag requires "Azure". 
You might need to introduce a new mapping function in internal/avs package to map plan ID to provider tag.

ebensom (changes requested):

internal/avs/external_eval_assistant.go
i332322: ah youre right. i was not thinking about this. aws and gcp it works but not for azure.

i332322:
currently this tagclass allows 3 provider like this. hence the code change to align with it.

image

i332322:
or the other way is make tag class value align with providerCodeByPlan: 
image

internal/avs/external_eval_assistant.go
i332322: ive made my changes.

ebensom (approved):

jaroslaw-pieszka (approved):

PR #193, opened 2023-11-21
Simple Time Measurements

ralikio:
**Description**

- Simple log message for making it easier to identify slow test parts.

**Related issue(s)**

### DIFF
--- a/internal/broker/poller.go
+++ b/internal/broker/poller.go
@@ -47,7 +47,7 @@ func (p *PassthroughPoller) Invoke(logic func() (bool, error)) error {
 type TimerPoller struct {
 	PollInterval time.Duration
 	PollTimeout  time.Duration
-	Log func(args ...any)
+	Log          func(args ...any)
 }
 
 func (p *TimerPoller) Invoke(logic func() (bool, error)) error {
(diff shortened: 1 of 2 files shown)


internal/broker/poller.go
```diff
@@ -43,3 +43,18 @@ func (p *PassthroughPoller) Invoke(logic func() (bool, error)) error {
...
+
+type TimerPoller struct {
+	PollInterval time.Duration
+	PollTimeout  time.Duration
+	Log func(args ...any)
+}
+
+func (p *TimerPoller) Invoke(logic func() (bool, error)) error {
+	var start = time.Now()
+	result := wait.PollImmediate(p.PollInterval, p.PollTimeout, func() (bool, error) {
```
jaroslaw-pieszka, line 55: `pollImmediate` is deprecated in latest Go version.

internal/broker/poller.go
```diff
@@ -43,3 +43,18 @@ func (p *PassthroughPoller) Invoke(logic func() (bool, error)) error {
...
+		return logic()
+	})
+	p.Log(fmt.Sprintf("Waiting for the status took: %v", time.Since(start)))
```
jaroslaw-pieszka, line 58: It is not easy to comprehend that we call `logic` to get `status`.  Either comment or more consistent names would be beneficial.

internal/broker/poller.go
ralikio: We decided not to do it right now.

internal/broker/poller.go
ralikio: Changed the message.

Issue #98, opened 2023-10-11
KEB release pipeline and Semantic Versioninig


PK85:
**Description**

We must provide a GA pipeline to release a new KEB version on the given tag. Then we can use it and bump in the resources directory.

**Reason**

Better release process than now.

**AC**
- [x] 1) take all the work we spend to create a release pipeline in the BTP-Manager, like:
  - [x] 1.1)  release notes generation
  - [x] 1.2) KEB new labels bug/enha/feature + checking them on PR level (same as in the BTP Manager)
- [x] 2) document new release process 
- [x] 3) rethink what kind of tests we use. Right now it is only unit tests. Think about how to test it more in the integration way. Result must be a decision what are the next steps.

**Links**
- info what SRE is doing around KEB testing: https://github.tools.sap/kyma/management-plane-charts/issues/354

piotrmiskiewicz:
- test-infra release job: https://github.com/kyma-project/test-infra/pull/9265 (only KEB)
-

MarekMichali:
- [ ] New release label validator

PR #235, opened 2023-12-01
Replace release label validation script

MarekMichali:
**Description**

Replace the existing bash label validation script with a python script. This new script will only be used during release. Validation at the PR level will still be done by the old bash script.

**Related issue(s)**

See also #98

### DIFF
--- a/scripts/python/release_label_validator.py
+++ b/scripts/python/release_label_validator.py
@@ -49,4 +49,4 @@ if invalid_prs:
     print('\n'.join([f"PR: {pr}" for pr in invalid_prs]))
     sys.exit(1) 
 
-print("\nAll PRs have exactly one label from the pool")
+print("\nAll PRs have exactly one required label")
--- a/scripts/python/release_label_validator.py
+++ b/scripts/python/release_label_validator.py
@@ -1,0 +1,54 @@
+import requests
+import os
+import yaml
+import sys
+
+with open('.github/release.yml', 'r') as file:
+    try:
+        release_yaml = yaml.safe_load(file)
+        label_pool = []
+        for category in release_yaml['changelog']['categories']:
+            label_pool.extend(category['labels'])
+    except yaml.YAMLError as exc:
+        print(exc)
+
+print(f"Label pool: {label_pool}")
+token = os.getenv('GITHUB_TOKEN')
+repo = os.getenv('REPOSITORY')
+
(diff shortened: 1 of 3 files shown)


scripts/python/release_label_validator.py
```diff
@@ -0,0 +1,54 @@
...
+invalid_prs = []
+for pr in prs_since_last_release:
+    labels = [label['name'] for label in pr['labels']]
+    common_labels = set(labels).intersection(label_pool)
+    if len(common_labels) != 1:
+        invalid_prs.append(pr['html_url'])
+    else:
+        valid_prs.append(pr['html_url'])
+
+print("\nThese PRs have exactly one label from the pool:")
```
jaroslaw-pieszka, line 43: `pool` term is kind of vague here.

scripts/python/release_label_validator.py
```diff
@@ -0,0 +1,54 @@
...
+for pr_url in valid_prs:
+    print(f"PR: {pr_url}")
+
+
+if invalid_prs:
+    print("\nThese PRs don't have exactly one label from the pool:")
+    for pr_url in invalid_prs:
```
jaroslaw-pieszka, line 50: We could join all references with a separator and print in separate line with newline at the end.

scripts/python/release_label_validator.py
```diff
@@ -0,0 +1,54 @@
...
+latest_release = response.json()
+latest_release_date = latest_release['created_at']
+
+response = requests.get(f'https://api.github.com/repos/{repo}/pulls?state=closed&sort=updated&direction=desc', headers={'Authorization': f'token {token}'})
+response.raise_for_status()
+all_closed_prs = response.json()
+
+prs_since_last_release = [
+    pr for pr in all_closed_prs
+    if pr['merged_at'] is not None and pr['merged_at'] > latest_release_date
```
jaroslaw-pieszka, line 30: How about these merged after the release was created but not taken by release.
I mean is it possible that merged_at > latest_release_date but the point in time when we created tag and cut off some PRs to be released is after latest_release_date.

scripts/python/release_label_validator.py
MarekMichali: Changed to `required labels`

jaroslaw-pieszka (approved):

PR #264, opened 2023-12-11
Remove deprecated components list provider and kyma-operator dependency

szwedm:
**Description**

- remove deprecated ComponentsListProvider,
- remove deprecated kyma-operator dependency,
- fix tests.

**Related issue(s)**


cmd/broker/testdata/kyma-components.yaml
```diff
@@ -1,5 +1,7 @@
 ---
 defaultNamespace: kyma-system
+prerequisites:
+  - name: "cluster-essentials"
 components:
   - name: "ory"
```
jaroslaw-pieszka, line 6: Why do we still use ory here?

cmd/broker/update_test.go
```diff
@@ -2393,7 +2395,8 @@ func TestUpdateBTPOperatorCredsSuccess(t *testing.T) {
 	i, err = suite.db.Instances().GetByID(id)
 	assert.NoError(t, err, "getting instance after update")
 	assert.ElementsMatch(t, rsu2.KymaConfig.Components, []*gqlschema.ComponentConfigurationInput{})
-	assert.ElementsMatch(t, componentNames(rsu2.ClusterSetup.KymaConfig.Components), []string{"ory", "monitoring", "btp-operator"})
+	assert.ElementsMatch(t, componentNames(rsu2.ClusterSetup.KymaConfig.Components),
+		[]string{"cluster-essentials", "ory", "monitoring", "btp-operator"})
```
jaroslaw-pieszka, line 2399: The same question - do we still need ory?

cmd/broker/testdata/kyma-components.yaml
szwedm: It doesn't really matter, tests don't provision anything as provisioner and reconciler are mocked. We just check if configuration sent from KEB matches the configuration other components receive. We can change the components to generic names, like `component1`, `component2`, but the required follow up is changing assertions where configurations are asserted.

cmd/broker/update_test.go
szwedm: We don't need ory, but for tests it doesn't matter. See my other answer: https://github.com/kyma-project/kyma-environment-broker/pull/264#discussion_r1422517080

jaroslaw-pieszka (approved):

Issue #71, opened 2023-09-27
New internal endpoint for Trial instances expiration


szwedm:
**Description**

The issue is an improvement request created after resolving https://github.com/kyma-project/control-plane/issues/3024. Kyma Environment Broker should have a new internal endpoint for handling the expiration of Trial instances. 

**Reasons**

New endpoint will encapsulate the instance expiration process. The process flow will be clear and serve just the purpose of expiring instances with Trial plan instead of mixing processes of the instance update and instance expiration, as it is now in the [instance update endpoint](https://github.com/kyma-project/kyma-environment-broker/blob/main/internal/broker/instance_update.go). The new endpoint can be set as an instance expiration request target in Trial cleanup Job.

**AC**
- [x] 1) New endpoint designed for expirations only, called only by trial-cleanup-cronjob. Adjust documentation about this job. Document how expiration works, and how active=true/false works. https://github.com/kyma-project/kyma-environment-broker/pull/142
- [x] 2) Find a way and create "e2e" tests for cronjob and new endpoint. If KEB or cronjob has a change then the test should be executed. https://github.com/kyma-project/kyma-environment-broker/pull/234 https://github.com/kyma-project/kyma-environment-broker/pull/242 https://github.com/kyma-project/kyma-environment-broker/pull/272 https://github.com/kyma-project/test-infra/pull/9524
- [x] 3) Cleanup existing update endpoint from expiration logic. ! We still need to keep some things about that ! https://github.com/kyma-project/kyma-environment-broker/pull/280
- [x] 4) ~~Include the new endpoint external request test in [KEB's endpoints test](https://github.com/kyma-project/kyma-environment-broker/tree/main/testing/e2e/skr/keb-endpoints-test).~~ NOT REQUIRED - new trial expiration endpoint is available only in the cluster through Service name https://github.com/kyma-project/kyma-environment-broker/pull/281 https://github.com/kyma-project/kyma-environment-broker/pull/285

szwedm:
## Trial expiration endpoint proposal

Sending a request triggers a Service Instance with Trial plan expiration. KEB sets `ERSContext.active` field to `false` and `expiredAt` to the the current local time (request processing start time). KEB creates a suspension operation to keep the record of the instance in the database until the instance's owner triggers the instance deprovisioning. The instance cannot be unsuspended.

### Request
#### Route 
`PUT /expire/service_instance/:instance_id`

[...]

PR #280, opened 2023-12-18
Remove trial expiration logic from update endpoint

szwedm:
**Description**

- remove expiration param processing from the update endpoint,
- enhance expiration endpoint with suspension operation id response for correct expiration request,
- clean tests.

**Related issue(s)**

See #71


cmd/broker/expiration_test.go
```diff
@@ -0,0 +1,315 @@
...
+		// when
+		resp = suite.CallAPI(http.MethodPut,
+			fmt.Sprintf(expirationRequestPathFormat, instanceID),
+			"")
+
+		// then
+		assert.Equal(t, http.StatusAccepted, resp.StatusCode)
+
+		suspensionOpID2 := suite.DecodeOperationID(resp)
+		assert.NotEmpty(t, suspensionOpID2)
```
jaroslaw-pieszka, line 160: Could we assert that suspensionOpID1 != suspensionOpID2?

cmd/broker/expiration_test.go
```diff
@@ -0,0 +1,315 @@
...
+
+		suite.WaitForOperationState(suspensionOpID2, domain.Succeeded)
+
+		actualInstance = suite.GetInstance(instanceID)
+		assert.True(t, actualInstance.IsExpired())
```
jaroslaw-pieszka, line 165: This code is duplicated: asserting that it's expired and active is false. 
How about extracting? 
I am not sure if this would make the code easier to read, but this idea crossed my mind.

jaroslaw-pieszka (dismissed):

cmd/broker/expiration_test.go
szwedm: sure, I will add the assertion.

cmd/broker/expiration_test.go
szwedm: suggestion applied.

jaroslaw-pieszka (approved):

piotrmiskiewicz (approved):

PR #288, opened 2023-12-19
Disable Duplicated Tests

ralikio:
**Description**

It looks like tag `database_integration` from `test_with_database_network` make target is not used anywhere, which results in duplicated invocation of tests that are also invoked in previous targets.

- Remove unnecessary tests invocation.

**Related issue(s)**


ralikio:
Useless PR...

PR #308, opened 2023-12-22
Use kubeconfigprovider

piotrmiskiewicz:
**Description**

- remove usage of provisioner to get the kubeconfig

There is only one exception - the provisioning process, because:
The code should work with and without infrastructure manager.

1. Infrastructure Manager is enabled:
KEB gathers kubeconfig and put it into the operation
KEB creates and waits for GardenerCluster readiness (when it is ready, a secret with kubeconfig is created by IM)
KEB tries to create the kubeconfig secret, but it is already created, so there is nothing to do.

2. Infrastructure Manager integration is disabled
KEB gathers kubeconfig and put it into the operation
KEB creates the kubeconfig secret (because there is no such secret)

then, whenever k8s client is needed, the common provider is used to read kubeconfig from the secret and create client.Client

**Related issue(s)**


internal/process/deprovisioning/btp_operator_cleanup.go
```diff
@@ -51,14 +49,10 @@ func (s *BTPOperatorCleanupStep) Name() string {
 	return "BTPOperator_Cleanup"
 }
 
-func (s *BTPOperatorCleanupStep) softDelete(operation internal.Operation, log logrus.FieldLogger) (internal.Operation, time.Duration, error) {
-	k8sClient, err := s.getKubeClient(operation, log)
-	if err != nil || k8sClient == nil {
-		return s.retryOnError(operation, err, log, "failed to get kube client")
-	}
+func (s *BTPOperatorCleanupStep) softDelete(operation internal.Operation, k8sClient client.Client, log logrus.FieldLogger) (internal.Operation, time.Duration, error) {
```
piotrmiskiewicz, line 52: softDelete has a client to SKR k8s as an argument

internal/process/deprovisioning/btp_operator_cleanup.go
```diff
@@ -81,15 +75,19 @@ func (s *BTPOperatorCleanupStep) softDelete(operation internal.Operation, log lo
...
 
 	if len(errors) != 0 {
-		return s.retryOnError(operation, fmt.Errorf(strings.Join(errors, ";")), log, "failed to cleanup")
+		return s.retryOnError(operation, nil, fmt.Errorf(strings.Join(errors, ";")), log, "failed to cleanup")
 	}
 	return operation, 0, nil
 }
 
 func (s *BTPOperatorCleanupStep) Run(operation internal.Operation, log logrus.FieldLogger) (internal.Operation, time.Duration, error) {
+	kclient, err := s.k8sClientProvider.K8sClientForRuntimeID(operation.RuntimeID)
```
piotrmiskiewicz, line 84: k8s client is gathered once, then used in other methods (passed as an argument)

internal/process/provisioning/get_kubeconfig_step.go
```diff
@@ -18,16 +16,13 @@ type GetKubeconfigStep struct {
 	provisionerClient   provisioner.Client
 	operationManager    *process.OperationManager
 	provisioningTimeout time.Duration
-	k8sClientProvider   func(kubeconfig string) (client.Client, error)
 }
 
 func NewGetKubeconfigStep(os storage.Operations,
```
piotrmiskiewicz, line 21: Get kubeconfig only gets kubconfig from provisioner and store it in the operation. No need to create (and store) k8s client. The client will be created by the provider when it will be needed

piotrmiskiewicz: This is the only place , where we call Provisioner because not all environments has enabled Infrastructure Manager. In case of Infrastructure Manager is disabled, KEB gets the kubeconfig and stores it in the secret

jaroslaw-pieszka (approved):

cmd/broker/provisioning.go
```diff
@@ -115,20 +115,20 @@ func NewProvisioningProcessingQueue(ctx context.Context, provisionManager *proce
 			step:      steps.NewCheckGardenerCluster(db.Operations(), cli),
 			condition: provisioning.SkipForOwnClusterPlan,
 		},
-		{
+		{ // TODO: this step must be removed when kubeconfig is created by IM
```
jaroslaw-pieszka, line 118:
```suggestion
		{ // TODO: this step must be removed when kubeconfig is created by IM only
```

jaroslaw-pieszka (approved):

PR #322, opened 2024-01-04
remove temporary featutre flag

ukff:


resources/keb/templates/deployment.yaml
```diff
@@ -69,8 +69,6 @@ spec:
               value: "{{ .Values.regionParameterIsRequired }}"
             - name: APP_BROKER_ALLOW_NETWORKING_PARAMETERS
               value: "{{ .Values.allowNetworkingParameters }}"
-            - name: APP_BROKER_ALLOW_MODULES_PARAMETERS
```
MarekMichali, line 72: This should be removed after KEB release with changes from this PR.

MarekMichali (approved):

PR #331, opened 2024-01-08
Remove allowNetworkingParameters feature flag

### LINKED ISSUE #324: Cleanup allowNetworkingParameters feature flag
This feature is enabled on every landscape so this flag is no longer needed.

KsaweryZietara:
**Related issue(s)**
Resolves #324


ralikio (approved):

resources/keb/templates/deployment.yaml
```diff
@@ -67,8 +67,6 @@ spec:
               value: "{{ .Values.exposeSchemaWithRegionRequired }}"
             - name: APP_BROKER_REGION_PARAMETER_IS_REQUIRED
               value: "{{ .Values.regionParameterIsRequired }}"
-            - name: APP_BROKER_ALLOW_NETWORKING_PARAMETERS
```
MarekMichali, line 70: This should be removed after KEB release with changes from this PR. After the release in the PR with the image bump, this can be removed. The reason is that the chart will be propagated without the env, but there will be an old KEB image expecting the `APP_BROKER_ALLOW_NETWORKING_PARAMETERS` flag.

MarekMichali (approved):

PR #334, opened 2024-01-09
Fix cmd/broker unit test flakiness 

piotrmiskiewicz:
**Description**

- sometimes update tests were failing because after the update (PATCH) the reconciler fake client was guided to a response for the previous operation (which has finished) and the fresh reconciler operation was never ended.

The change was to program the fake client with a response before any interaction. 

**Related issue(s)**

### DIFF
--- a/internal/reconciler/fake.go
+++ b/internal/reconciler/fake.go
@@ -1,11 +1,11 @@
 package reconciler
 
 import (
-    "fmt"
-    "sync"
-    "time"
+	"fmt"
+	"sync"
+	"time"
 
-    reconcilerApi "github.com/kyma-incubator/reconciler/pkg/keb"
+	reconcilerApi "github.com/kyma-incubator/reconciler/pkg/keb"
 )
 
 /*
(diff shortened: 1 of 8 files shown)


internal/reconciler/fake.go
```diff
@@ -61,15 +70,40 @@ func (c *FakeClient) GetCluster(clusterName string, configVersion int64) (*recon
...
+		msg := intKeys(existingCluster.clusterStates)
+		return &reconcilerApi.HTTPClusterResponse{}, fmt.Errorf("cluster state not found, version=%d (%s)", configVersion, msg)
+	}
+	if existingCluster.numberOfGetStatusCalls == 0 && c.expectedStatus != nil {
+		state.Status = *c.expectedStatus
 	}
 	return state, nil
 }
 
+func intKeys(states map[int64]*reconcilerApi.HTTPClusterResponse) string {
```
jaroslaw-pieszka, line 91: Do we just join all states from states map?

internal/reconciler/fake.go
```diff
@@ -61,15 +70,40 @@ func (c *FakeClient) GetCluster(clusterName string, configVersion int64) (*recon
...
+	result := ""
+	for k, _ := range states {
+		result = result + fmt.Sprintf("[%d] ", k)
+	}
+	return result
+}
+
+func stringKeys(clusters map[string]*registeredCluster) string {
```
jaroslaw-pieszka, line 99: And do we here join all clusters from inventoryClusters?

internal/reconciler/fake.go
piotrmiskiewicz: I've added states

jaroslaw-pieszka (approved):

PR #382, opened 2024-01-22
Kyma resource name is always as the first runtimeID

### LINKED ISSUE #363: Kyma resource name is always as the first runtimeID
**Description**

The Kyma resource name is equal to runtimeID. The problem is that when trial SKR is suspended, then unsuspension does not create a new name for Kyma resource (runtimeID has been changed).

**AC**
- [x] 1) first check if GardenerCluster CR has the same issue or not
- [x] 2) apply solution

**Solution**

Solution: KymaResourceName field must be set empty in the deprovisioning process.
Consider a proper e2e  test in cmd/broker folder.

KsaweryZietara:
**Related issue(s)**
Resolves #363


cmd/broker/update_test.go
```diff
@@ -405,6 +405,79 @@ func TestUpdateContext(t *testing.T) {
...
+           "globalaccount_id": "g-account-id",
+           "user_id": "john.smith@email.com",
+			"active": true
+       }
+       
+   }`)
+	assert.Equal(t, http.StatusOK, resp.StatusCode)
+	suite.processProvisioningByInstanceID(iid)
+
+	suite.AssertKymaResourceExists(opID)
```
piotrmiskiewicz, line 475: this should fail, let's discuss it

internal/process/deprovisioning/remove_instance_step.go
```diff
@@ -57,6 +57,8 @@ func (s *RemoveInstanceStep) Run(operation internal.Operation, log logrus.FieldL
 		log.Info("Removing the RuntimeID field from the operation")
 		operation, backoff, _ = s.operationManager.UpdateOperation(operation, func(operation *internal.Operation) {
 			operation.RuntimeID = ""
+			operation.GardenerClusterName = ""
```
ralikio, line 60: Why not to do it in [handlers.go ](https://github.com/kyma-project/kyma-environment-broker/blob/eb0393a7df82aa60803126f5d704edad89b4448c/internal/suspension/handler.go#L145)? It is more suitable because that is were new operation is initialised. The current approach means that we are changing something for the process that resides in other places which might be harder to understand.

ralikio (changes requested):

piotrmiskiewicz (approved):

ralikio (changes requested):
Checking tests execution.

ralikio (approved):

PR #389, opened 2024-01-23
Runtime Operations Doc Correction

ralikio:
**Description**

- overview enhancement,
- changed content and placement of `Stage` section,
- replaced `SetOverrides` with `AppendOverrides`,
- added content about extending `InstanceDetails`,
- moved sentence about returning operation instance from `Run` method,
- removed examples for adding steps for update operation, added note about its similarity to provisioning examples.

**Related issue(s)**

### DIFF
--- a/docs/user/03-20-runtime-operations.md
+++ b/docs/user/03-20-runtime-operations.md
@@ -11,7 +11,7 @@ Kyma Environment Broker (KEB) allows you to configure operations you can run on 
 
 A stage is a grouping unit for steps. A step is a part of a stage. An operation can consist of multiple stages, and a stage can consist of multiple steps.  Once all the steps in a stage are successfully executed, the stage is marked as finished and never repeated again, even if the next one fails. If any steps fail at a given stage, the whole stage is repeated from the beginning.
 
-## Provision
+## Provisioning
 
 The provisioning process is executed when the instance is created, or an unsuspension is triggered.
 Each provisioning step is responsible for a separate part of preparing Kyma runtime. For example, in a step you can provide tokens, credentials, or URLs to integrate SAP BTP, Kyma runtime with external systems.
--- a/docs/user/03-20-runtime-operations.md
+++ b/docs/user/03-20-runtime-operations.md
@@ -1,11 +1,15 @@
 # SAP BTP, Kyma Runtime Operations
 
-Kyma Environment Broker (KEB) allows you to configure operations that you can run on a SAP BTP, Kyma runtime. Each operation consists of several steps and each step is represented by a separate file. As every step can be re-launched multiple times, for each step, you should determine a behavior in case of a processing failure. It can either:
+Kyma Environment Broker (KEB) allows you to configure operations that you can run on a SAP BTP, Kyma runtime. Each operation is processed by several steps arranged in stages and ordered in queue structure. During its processing, an operation is passed to every step for it to act opon it acording to its responsibility. Steps get to process operation in order of its arrangement in the queue. As every step can be re-launched multiple times, for each step, you should determine a behavior in case of a processing failure. It can either:
 
-- Return an error, which interrupts the entire process, or
-- Repeat the entire operation after the specified period.
+- Return an error, which interrupts the entire process, or skips step execution.
+- Repeat the entire operation after specified period.
 
 > **NOTE:** It's important to set lower timeouts for the Kyma installation in the Runtime Provisioner.
+
+## Stages
+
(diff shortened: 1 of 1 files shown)


docs/user/03-20-runtime-operations.md
```diff
@@ -1,12 +1,16 @@
...
-- Return an error, which interrupts the entire process, or
-- Repeat the entire operation after the specified period.
+- Return an error, which interrupts the entire process, or skips step execution.
+- Repeat the entire operation after specified period.
 
 > **NOTE:** It's important to set lower timeouts for the Kyma installation in the Runtime Provisioner.
 
+## Stages
+
+A stage is a grouping unit for steps. A step is a part of a stage. An operation can consist of multiple stages, and a stage can consist of multiple steps.  Once all the steps in a stage are successfully executed, the stage is marked as finished and never repeated again, even if the next one fails. If any steps fail at a given stage, the whole stage is repeated from the beginning.
```
IwonaLanger, line 12:
```suggestion
A stage is a grouping unit for steps. An operation can consist of multiple stages, and a stage can consist of multiple steps. Once all the steps in a stage are successfully executed, the stage is marked as finished and never repeated, even if the next stage fails. If a step fails at a given stage, the whole stage is repeated from the beginning.
```

docs/user/03-20-runtime-operations.md
```diff
@@ -1,12 +1,16 @@
...
+
 ## Provision
```
IwonaLanger, line 14:
```suggestion
## Provisioning
```

docs/user/03-20-runtime-operations.md
```diff
@@ -43,7 +47,7 @@ You can find all the upgrading Kyma steps in the [upgrade_kyma](../../cmd/broker
 
 ## Provide Additional Steps
 
-You can configure SAP BTP, Kyma runtime operations by providing additional steps. To add a new step, follow these tutorials:
+You can configure SAP BTP, Kyma runtime operations by providing additional steps. Every operation (see implementation of `internal.Operation` structure) is based on the same Operation structure. The following examples presents how to extend KEB process based on provisioning operation. Extensions for other processes follow the same steps but just require its specific structures.
```
IwonaLanger, line 50:
```suggestion
You can configure SAP BTP, Kyma runtime operations by providing additional steps. Every operation (see the implementation of `internal.Operation` structure) is based on the same Operation structure. The following examples present how to extend the KEB process based on provisioning operation. Extensions for other processes follow the same steps but require their specific structures.
```

IwonaLanger: Where or how can I `see implementation of `internal.Operation` structure`?

docs/user/03-20-runtime-operations.md
```diff
@@ -63,11 +67,10 @@ You can configure SAP BTP, Kyma runtime operations by providing additional steps
     ```
 
    - `Name()` method returns the name of the step that is used in logs.
-   - `Run()` method implements the functionality of the step. The method receives operations as an argument to which it can add appropriate overrides or save other used variables.
-
+   - `Run()` method implements the functionality of the step. The method receives operations as an argument to which it can add appropriate overrides or save other used variables. You should always return the modified operation from the method.
```
IwonaLanger, line 70:
```suggestion
   - `Run()` method implements the functionality of the step. The method receives operations as an argument to which it can add appropriate overrides or save other used variables. You must always return the modified operation from the method.
```

docs/user/03-20-runtime-operations.md
```diff
@@ -80,9 +83,7 @@ You can configure SAP BTP, Kyma runtime operations by providing additional steps
     })
     ```
 
-    If your functionality contains long-term processes, you can store data in the storage.
-    To do this, add the following field to the provisioning operation in which you want to save data:
-
+    If your functionality requires saving data in the storage, you can add this by adding fields to the generic internal.Operation, specific implementation of that structure or the InstanceDetails, all of which are defined in [model.go]](../../internal/model.go). The difference is that for specific operation implementation new fields are only visible for that specific type and InstanceDetails is copied during operation initialization across all operations that concert given runtime. The example below shows how to extend operation with additional fields:
```
IwonaLanger, line 86:
```suggestion
    If your functionality requires saving data in the storage, you can do it by adding fields to the generic `internal.Operation`, a specific implementation of that structure, or the InstanceDetails, all of which are defined in [model.go](../../internal/model.go). The difference is that for a specific operation implementation, new fields are only visible for that specific type and InstanceDetails is copied during operation initialization across all operations that concert a given runtime. The example below shows how to extend operation with additional fields:
```

IwonaLanger (approved):

PR #426, opened 2024-02-05
Update istio gwateway config

berkanercan:
**Description**

- SRE Team has introduced global config values for istio gateway. In that regard, we are updating the helm templates to use that global value. 
- ...
- ...

**Related issue(s)**


gczuczy (approved):

szwedm (approved):
/lgtm

PR #427, opened 2024-02-06
Fix Gardener cluster name and kyma name cleanup in the deprovisioning…

piotrmiskiewicz:
… process

**Description**
The Bug is if the delete gardener cluster resource fails, the name of the resource will be cleaned so it will never 

- GardenerClusterName and KymaResourceName are cleaned if it was really removed

Because of orphaned resources (GardenerCluster and Kyma) - the cleaning steps takes runtime ID from the instance and use it if the saved resource name is empty.

**Related issue(s)**

### DIFF
--- a/internal/process/deprovisioning/delete_kyma_resource_step.go
+++ b/internal/process/deprovisioning/delete_kyma_resource_step.go
@@ -30,14 +30,16 @@ type DeleteKymaResourceStep struct {
 	kcpClient          client.Client
 	configProvider     input.ConfigurationProvider
 	defaultKymaVersion string
+	instances          storage.Instances
 }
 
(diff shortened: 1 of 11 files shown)


internal/process/deprovisioning/delete_kyma_resource_step.go
```diff
@@ -71,7 +73,25 @@ func (step *DeleteKymaResourceStep) Run(operation internal.Operation, logger log
 	}
 	kymaResourceName := steps.KymaName(operation)
 	if kymaResourceName == "" {
-		logger.Infof("Kyma resource name is empty, skipping")
+		logger.Infof("Kyma resource name is empty, using isntance.RuntimeID")
```
jaroslaw-pieszka, line 76:
```suggestion
		logger.Infof("Kyma resource name is empty, using instance.RuntimeID")
```

jaroslaw-pieszka (approved):

PR #507, opened 2024-03-05
Index on created_at Field

ralikio:
**Description**

- Index on instance's `created_at` field to optimize list instances [query](https://github.com/kyma-project/kyma-environment-broker/blob/dc04e0d709a16b267e168f556d794d06cef6b8ea/internal/storage/postsql/read.go#L670).

**Related issue(s)**

### DIFF
--- a/resources/keb/migrations/202403051613_add_indexes.up.sql
+++ b/resources/keb/migrations/202403051613_add_indexes.up.sql
@@ -1,0 +1,1 @@
+CREATE INDEX instances_by_created_at ON instances USING btree (created_at);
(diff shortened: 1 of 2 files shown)


resources/keb/migrations/202403051613_add_indexes.up.sql
```diff
@@ -0,0 +1 @@
+CREATE INDEX instances_by_created_at ON instances USING btree (created_at);
```
jaroslaw-pieszka, line 1: Any. rationale why index for `instances` helps?
Could you attach at least explain without and with this index created?

piotrmiskiewicz: the `list` query has `order by instances.created_at` part

piotrmiskiewicz (approved):

jaroslaw-pieszka (approved):

PR #514, opened 2024-03-06
Query Optimisation - Group Operations in Go

ralikio:
PR with grouping multiple database queries for loading operation by instance id and type into a single query that loads all instance operations and groups them by type in the code.


internal/storage/driver/memory/operation.go
```diff
@@ -190,6 +190,46 @@ func (s *operations) ListOperationsByInstanceID(instanceID string) ([]internal.O
 	return operations, nil
 }
 
+func (s *operations) ListOperationsByInstanceIDGroupByType(instanceID string) (*internal.GroupedOperations, error) {
```
jaroslaw-pieszka, line 193: Is this really faster when we fetch all and group and sort in memory?

internal/storage/driver/postsql/operation.go
```diff
@@ -169,6 +169,61 @@ func (s *operations) ListOperationsByInstanceID(instanceID string) ([]internal.O
 	return ret, nil
 }
 
+func (s *operations) ListOperationsByInstanceIDGroupByType(instanceID string) (*internal.GroupedOperations, error) {
```
jaroslaw-pieszka, line 172: Is this really faster when we fetch all and group and sort in memory?

jaroslaw-pieszka: So if we fetch operations for particular instanceID this approach make sense.
But where do we need such data?
Operations per instanceID grouped by type?

piotrmiskiewicz: the response of runtimes endpoint groups operations by type

internal/storage/driver/memory/operation.go
```diff
@@ -190,7 +190,47 @@ func (s *operations) ListOperationsByInstanceID(instanceID string) ([]internal.O
...
+
+		case internal.OperationTypeUpgradeKyma:
+			grouped.UpgradeKymaOperations = append(grouped.UpgradeKymaOperations, internal.UpgradeKymaOperation{Operation: op})
+
+		case internal.OperationTypeUpgradeCluster:
+			grouped.UpgradeClusterOperations = append(grouped.UpgradeClusterOperations, internal.UpgradeClusterOperation{Operation: op})
+
+		case internal.OperationTypeUpdate:
+			grouped.UpdateOperations = append(grouped.UpdateOperations, internal.UpdatingOperation{Operation: op})
+		}
```
jaroslaw-pieszka, line 221: Could we log if there is type not matching any of these?

internal/storage/driver/memory/operation.go
```diff
@@ -190,7 +190,47 @@ func (s *operations) ListOperationsByInstanceID(instanceID string) ([]internal.O
...
+	}
+
+	s.sortProvisioningByCreatedAtDesc(grouped.ProvisionOperations)
```
jaroslaw-pieszka, line 224: If the input is sorted by CreatedAt the sort here would not be necessary. Operations are already sorted.

ralikio: Because it is in memory implementation the sorting order is not guaranteed and heavily dependent on insertion order in specific test.

internal/storage/driver/memory/operation.go
ralikio: Corrected in https://github.com/kyma-project/kyma-environment-broker/pull/514/commits/3b4a8b8200c665f42373a2dd3172cd17c09df5b5. I have added an error for postgresql DAO and panic for in memory implementation.

PR #541, opened 2024-03-11
Ensure inputCreator is not nil

### LINKED ISSUE #492: [TO VERIFY] Ensure inputCreator is not nil and there is no panic in KEB causing pod to restart
We shall make our code and configuration more secure in regard to possibility of panic in KEB.

**AC**
- [ ] avoid overwriting of in-memory operations data by data from storage when handling update conflict (operations_manager.go: `UpdateOperation`) - currently in case of conflict we get operation from the storage so all data in storage structure are overwritten by data from persistent memory, e.g. inputCreator which is never serialized is overwritten with `nil` value
- [x] handle panic in a way that we avoid pod restart
- [ ] use nil checks for inputCreator to minimize risk of panic
- [ ] verify KEB pod strategy to make sure in no scenario there are more than single KEB pod running - this is the possible reason of conflict while updating operation in persistent storage

KsaweryZietara:
**Description**

- merge persistent operation with the in memory operation when handling update conflict,
- remove unused files. 

**Related issue(s)**
Resolves #492


piotrmiskiewicz:
we need to be sure what to do if the input creator is nil.... what KEB should do with the operation

piotrmiskiewicz (approved):

PR #550, opened 2024-03-13
adding new oidc group

i332322:
related task: https://github.tools.sap/kyma/backlog/issues/5018

### DIFF
--- a/resources/keb/templates/authorization-policy.yaml
+++ b/resources/keb/templates/authorization-policy.yaml
@@ -151,6 +151,7 @@ spec:
       values:
       - {{ .Values.oidc.groups.admin }}
       - {{ .Values.oidc.groups.operator }}
+      - {{ .Values.oidc.groups.viewer }}
   - to:
     - operation:
         methods:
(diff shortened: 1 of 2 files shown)


i332322:
kind/enhancement

resources/keb/templates/authorization-policy.yaml
```diff
@@ -151,6 +151,7 @@ spec:
       values:
       - {{ .Values.oidc.groups.admin }}
       - {{ .Values.oidc.groups.operator }}
+      - {{ .Values.oidc.groups.viewer }}
```
gczuczy, line 154: I think having having a generic option for extending it would be better. something like a `.Values.oidc.extraGroups`, and just append the list as it is.

gczuczy (changes requested):

resources/keb/templates/authorization-policy.yaml
i332322: after discussing with Gergely, this approach is also acceptable.

ebensom (approved):

Issue #60, opened 2023-09-20
[betaEnabled] Support Beta flag in KEB and Kyma CR


PK85:
**Description**

The KEB needs to store “Used for production” and “Enabled beta features” information during Kyma provisioning (Only provisioning).

**Reason**

This feature is of high importance as it directly impacts the ability of customers to utilize beta modules effectively within Kyma. Being able to discern user preferences for beta features during Subaccount creation will enable better alignment with customer needs and preferences.

**AC**
- [x] 0) do local POC with all proper working calla to CIS APIs and present to the team, then next points:
~~- [ ] 1) When KEB is called with `ers_context` structure we need to take subaccountID and find  “Enabled beta features” information in the Account Service API. Then KEB persists that in the DB and labels `Kyma` CR during Kyma provisioning [EDIT: not valid any more]~~
- [x] 2) Be able to check when beta flag has been change on our side
- [x] 3) figure out how to test that new feature 
- [x] 4) Update all Runtimes created in the past that have missing beta information.
  - [x] 4.1) run it at least once
- [x] 5) Provide documentation about new integration

**Info**

- Docs about Cloud Management Service (CIS) APIs: 
  - https://accounts-service.cfapps.eu10.hana.ondemand.com/api?urls.primaryName=Accounts-Service-Api$cis-system&scope=cis-system#/g_Accounts%20APIs/getSubaccount_1
  - https://help.sap.com/docs/btp/sap-business-technology-platform/sap-cloud-management-service-service-plans

PK85:
We encountered that "Enable beta features" could be selected to "true" after subaccount creation. This change a way how we can implement this feature.

The idea to validate:
- looks for "Subaccount_Update" event type from Event Service API, then if "subaccountId" matches any of our Runtimes(right now it should be max exactly 1) do a call to AccountService API to check is beta has changed.

jaroslaw-pieszka:
Having **subaccount** with Cloud Management Service (cis) and `system-basic` plan you can get `Subaccount_Update` event when `betaEnabled` or `usedForProduction` has changed.
However in the event payload there is not information about `usedForProduction` even if change of this value triggered appropriate event.

betaEnabled and usedForProduction checked in cockpit - event generated:
```json
"events": [
        {
                "id": 25290915,
                "actionTime": 1706600916546,
                "creationTime": 1706600917056,
                "details": {
                        "description": "Subaccount updated.",
[...]

PR #574, opened 2024-03-18
Priority queue with callbacks

jaroslaw-pieszka:
**Description**

- queue with callbacks in order to use these for metrics gathering,
- queue element wrapped with data tracking time spend in the queue
- extracting with boolean flag to allow multiple consumers 

**Related issue(s)**

#60


internal/syncqueues/priority_queue_with_stats.go
```diff
@@ -0,0 +1,148 @@
...
+	q.elements[q.size] = e
+	q.size++
+	q.siftUp()
+
+	if q.eventHandler != nil && q.eventHandler.OnInsert != nil {
+		q.eventHandler.OnInsert(q.size)
+	}
+}
+
+func (q *SubaccountAwarePriorityQueueWithCallbacks) Extract() QueueElement {
```
piotrmiskiewicz, line 79: maybe Extract should return a bool which indicates if there was an item or not

piotrmiskiewicz:
do not log as error, when there is no item

piotrmiskiewicz:
do not log as INFO, it is at most DEBUG

internal/syncqueues/priority_queue_with_stats.go
jaroslaw-pieszka: Changes pushed

piotrmiskiewicz (approved):

Issue #593, opened 2024-03-27
Free plan - kyma for 30 days, one provisioning in the given Global Account


PK85:
**Description**

From: 
- Right now we offer one Kyma instance in the Global Account for free (plan "free"). 

To: 
- 30 days after creationTimestamp KEB must deprovision Kyma instance, 
- Another "free" plan provisioing in this GA must fail (We allow to provision only once in given GA)

**AC**
- [x] 1) Functionality:
  - [x] 1.0) Extend InstanceArchived table with status of the provisioning (we will need that information to handle a case where user provisioned first time and KEB failed, we need to allow him to provision again) 
PR: https://github.com/kyma-project/kyma-environment-broker/pull/599
  - [x] 1.1) Functionality available under feature flag PR: https://github.com/kyma-project/kyma-environment-broker/pull/600
  - [x] 1.2) make 30 days as configuration for job and for KEB (days left calculation)
PRs: [job](https://github.com/kyma-project/kyma-environment-broker/blob/main/cmd/expirator/main.go#L30), [KEB](https://github.com/kyma-project/kyma-environment-broker/blob/main/internal/broker/broker.go#L46)
  - [x] 1.3) Show with metadata label under service instance how many days left for expiration
PRs: https://github.com/kyma-project/kyma-environment-broker/pull/633 [#5103](https://secret.repo/kyma/management-plane-config/pull/5103)
  - [x] 1.4) When time comes, deprovision Kyma instance 
PRs: https://github.com/kyma-project/kyma-environment-broker/pull/602 https://github.com/kyma-project/kyma-environment-broker/pull/607
  - [x] 1.5) Do not allow to provision Kyma with "free" plan in the Global Account (if free was already consumed) 
PR: https://github.com/kyma-project/kyma-environment-broker/pull/608
  - [x] 1.6) We need to make sure to be able to run skr-free-aws-integration-dev and argocd promotion tests for free plan (idea: whitelist GA ids + flag "freeOnlyOnce:" true/false)
  PR: [#5097](https://secret.repo/kyma/management-plane-config/pull/5097)
- [x] 2) Documentation: (@IwonaLanger)
  - [x] 2.1) help sap documentation [free plan description](https://help.sap.com/docs/btp/sap-business-technology-platform/available-plans-in-kyma-environment?version=Cloud#free) updated
     - [publication requested ](https://jira.tools.sap/browse/NGPDOC-4714) 
[...]

PR #608, opened 2024-03-29
One freemium per global account

KsaweryZietara:
**Description**

- do not allow to provision Kyma with "free" plan in the Global Account (if free was already consumed),
- add tests.

**Related issue(s)**
See also #593


internal/broker/instance_create.go
```diff
@@ -338,6 +343,33 @@ func (b *ProvisionEndpoint) validateAndExtract(details domain.ProvisionDetails,
...
+			logger.Info("Provisioning Free SKR rejected, such instance was already created for this Global Account")
+			return ersContext, parameters, fmt.Errorf("free Kyma was created for the global account, but there is only one allowed")
+		}
+
+		filter := dbmodel.InstanceFilter{
+			GlobalAccountIDs: []string{ersContext.GlobalAccountID},
+			PlanIDs:          []string{FreemiumPlanID},
+			States:           []dbmodel.InstanceState{dbmodel.InstanceSucceeded},
+		}
+		_, _, count, err = b.instanceStorage.List(filter)
```
piotrmiskiewicz, line 362: currently, the instances archived is empty (there is no data). After a deprovisioning, the instance is deleted - this ios the reason we have to check operations instead of instances. Check, if there is an operation "provisioning" which has seuceeded for such global maccount

piotrmiskiewicz (approved):

piotrmiskiewicz (approved):

Issue #497, opened 2023-04-12
Timestamps inconsistent across sources (kibana, logs explorer, events table)


jaroslaw-pieszka:
While investigating one can use multiple sources of information but it is very hard to order log messages and events (from `events` table chronologically, since timestamps differ in value and accuracy.

Execution of Initialization step for instance_id: 096B7C10-78F4-4635-B6EE-F200502CAE89 message: "create provisioner input creator for "4deee563-e5ec-4731-b9b1-53b42d855f0c" plan ID"

- Logs Explorer - timestamp:` 2023-04-03T13:17:27Z`, receiveTimestamp: `2023-04-03T13:17:28.302456543Z`, 

- kibana -  timestamp `Apr 3, 2023 @ 13:17:27.341`

For second (possibly parallel) execution of Initialization step:

- Logs Explorer - timestamp: `2023-04-03T13:17:27Z`, receiveTimestamp: `2023-04-03T13:17:28.302456543Z`, 

- kibana -  timestamp `Apr 3, 2023 @ 13:17:32.696

Events:
```
  {
    "ID": "a0eb3768-d9b6-4d9d-97ef-4214db852f41",
    "Level": "info",
    "InstanceID": "096B7C10-78F4-4635-B6EE-F200502CAE89",
    "OperationID": "df044c63-f3d1-4238-8984-90ba4a04a4ba",
    "Message": "processing step: Provision_Initialization",
    "CreatedAt": "2023-04-03T13:17:27.225448Z"
  },
  {
    "ID": "f7ae3796-e149-4af8-9cf4-367e90c43bf0",
    "Level": "info",
    "InstanceID": "096B7C10-78F4-4635-B6EE-F200502CAE89",
    "OperationID": "df044c63-f3d1-4238-8984-90ba4a04a4ba",
    "Message": "processing step: Provision_Initialization",
    "CreatedAt": "2023-04-03T13:17:27.403207Z"
  },
```

Issues:
- in Logs Explorer timestamp is truncated to the seconds, and logs are not ordered chronologically
- timestamps differ across kibana, logs explorer and `events` table (even taking into consideration that event is persisted earlier than log message is printed)

ralikio:
Draft PR: https://github.com/kyma-project/kyma-environment-broker/pull/597

ralikio:
Timestamps corrected. Will be applied in the next release.

ralikio:
[...]

PR #654, opened 2024-04-11
Removed Println Invocations

ralikio:
**Description**

- Removed Invocations of Println and changed them to logrus calls.

**Related Issues**

https://github.com/kyma-project/kyma-environment-broker/issues/497


common/setup/builder.go
```diff
@@ -91,7 +91,11 @@ func (b *AppBuilder) WithBrokerClient() {
 }
 
 func (b *AppBuilder) WithProvisionerClient() {
-	b.provisionerClient = provisioner.NewProvisionerClient(b.cfg.Provisioner.URL, b.cfg.Provisioner.QueryDumping)
+	if b.logger == nil {
+		panic("Logger is not defined")
```
jaroslaw-pieszka, line 95: Is it really the reason to panic?
Needs discussion.

internal/model.go
```diff
@@ -736,9 +737,9 @@ type ConfigForPlan struct {
 	KymaTemplate         string          `json:"kyma-template" yaml:"kyma-template"`
 }
 
-func (c *ConfigForPlan) ContainsAdditionalComponent(componentName string) bool {
+func (c *ConfigForPlan) ContainsAdditionalComponent(componentName string, log logrus.FieldLogger) bool {
 	for _, c := range c.AdditionalComponents {
-		fmt.Println(c.Name)
+		log.Infof("looking for additional component %s", c.Name)
```
jaroslaw-pieszka, line 742: I think logging here is not necessary at all. 
I would just remove Println.

internal/process/provisioning/btp_operator_overrides.go
```diff
@@ -29,7 +29,7 @@ func (s *BTPOperatorOverridesStep) Run(operation internal.Operation, log logrus.
 	f := func(op *internal.Operation) {
 		op.InstanceDetails.ServiceManagerClusterID = clusterID
 	}
-	if !operation.InputCreator.Configuration().ContainsAdditionalComponent(internal.BTPOperatorComponentName) {
+	if !operation.InputCreator.Configuration().ContainsAdditionalComponent(internal.BTPOperatorComponentName, log) {
```
jaroslaw-pieszka, line 32: Will this part be removed soon? If so, I would not bother here.

jaroslaw-pieszka, line 32: But if this part is going to be left in code we should be consistent.
So we should pass `log` or `logger` but not both at the same time.

ralikio: Removed.

common/setup/builder.go
ralikio: I would leave it as it is should never be the case that logger is not defined.

jaroslaw-pieszka: I see, but the thing is that you have method WithLogger, and you allow not to call it.
This building method is optional, so if we require logger to be defined, we should make it mandatory.

ralikio: I have moved logger instance creation to builder constructor to make it implicit, ok? https://github.com/kyma-project/kyma-environment-broker/pull/654/commits/475c13b5e1d4336fabcdf019c0dd1654e88cc98f

piotrmiskiewicz (approved):

PR #669, opened 2024-04-15
Uniform Logger

ralikio:
**Description**

KEB code base uses multiple loggers. The following PR changes logger invocations to either slog or Loggrus that log messages as json messages (with custom handler set in the PR) and details timestamp (default in slog and customised in the PR for Loggrus).

- corrected order of Logrus initialisation in main methods,
- removed `log` package usages,
- set missing json handler for slog and default logrus loggers.

**Related issue(s)**

### DIFF
--- a/cmd/broker/main.go
+++ b/cmd/broker/main.go
@@ -461,7 +461,7 @@ func main() {
 		logs.Infof("Call handled: method=%s url=%s statusCode=%d size=%d", params.Request.Method, params.URL.Path, params.StatusCode, params.Size)
 	})
 
-	fatalOnError(http.ListenAndServe(cfg.Host+":"+cfg.Port, svr))
+	fatalOnError(http.ListenAndServe(cfg.Host+":"+cfg.Port, svr), logs)
 }
 
 func checkDefaultVersions(versions ...string) error {
@@ -643,7 +643,7 @@ func initClient(cfg *rest.Config) (client.Client, error) {
 	return cli, nil
 }
 
-func fatalOnError(err error, log *logrus.Logger) {
+func fatalOnError(err error, log logrus.FieldLogger) {
 	if err != nil {
 		log.Fatal(err)
 	}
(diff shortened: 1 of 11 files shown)


cmd/broker/main.go
```diff
@@ -228,10 +227,24 @@ func main() {
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
+	// set default formatted
+	logrus.SetFormatter(&logrus.JSONFormatter{
+		TimestampFormat: time.RFC3339Nano,
+	})
+	logs := logrus.New()
+	logs.SetFormatter(&logrus.JSONFormatter{
```
jaroslaw-pieszka, line 235: Code duplication. Why this is necessary to repeat the sentence. If we SetFormatter for `logrus` and then create new `log` would it not be set to json already?

cmd/broker/main.go
```diff
@@ -314,33 +318,33 @@ func main() {
 		kebConfig.NewConfigMapConverter())
 	componentsProvider := runtime.NewComponentsProvider()
 	gardenerClusterConfig, err := gardener.NewGardenerClusterConfig(cfg.Gardener.KubeconfigPath)
-	fatalOnError(err)
+	fatalOnError(err, logs)
```
jaroslaw-pieszka, line 321: Is this the only way, to pass always logger to this function? No way to use some `global` logger?

cmd/broker/main.go
ralikio: In the implementation there is one global instance of logrus that we need to configure for all the places where explicit logrus instance has not been created. Additionally, `logrus.New` does not seem to copy configuration from that global instance when creating a new one. That is why duplicated configuration is required.

cmd/broker/main.go
ralikio: I did it as a safe guard to use logger that is passed to methods in which `fataOnError` is invoked in. This way we are keeping all the labels that fataOnError should log.

jaroslaw-pieszka (approved):

Issue #647, opened 2024-04-10
Make Kyma resource deletion timeout configurable (set to 30s)


PK85:
**Description**

One of the steps in the deprovisioning process checks whether Kyma CR has been deleted. It does so in a loop with a fixed timeout. Allow to configure the check timeout with env value.

**Reason**
If any module connot be uninstalled because of blocking CRs we currently waste 30 min.

**AC**
- [x] 1) Timeout value is configurable through environment variable.
https://github.com/kyma-project/kyma-environment-broker/pull/672, [#5144](https://secret.repo/kyma/management-plane-config/pull/5144)
- [x] 2) Before going with new settings to stage and prod, set 30 sec on dev, create Kyma cluster, create instance via btp-operator module, desprovision Kyma and see if everything works as expected
- [x] 3) Set variable for every environment (separately) to 30 seconds [#5151](https://secret.repo/kyma/management-plane-config/pull/5151)

PR #672, opened 2024-04-17
Make Kyma resource deletion timeout configurable

KsaweryZietara:
**Description**

- set Kyma resource deletion timeout to 30 seconds,
- adjust tests.

**Related issue(s)**
See #647

### DIFF
--- a/internal/process/deprovisioning/check_kyma_resource_deleted_step.go
+++ b/internal/process/deprovisioning/check_kyma_resource_deleted_step.go
@@ -62,7 +62,7 @@ func (step *CheckKymaResourceDeletedStep) Run(operation internal.Operation, logg
 
 	if err == nil {
 		logger.Infof("Kyma resource still exists")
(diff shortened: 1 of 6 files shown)


internal/process/deprovisioning/check_kyma_resource_deleted_step.go
```diff
@@ -60,7 +62,7 @@ func (step *CheckKymaResourceDeletedStep) Run(operation internal.Operation, logg
 
 	if err == nil {
 		logger.Infof("Kyma resource still exists")
-		return step.operationManager.RetryOperationWithoutFail(operation, step.Name(), "Kyma resource still exists", 15*time.Second, 30*time.Minute, logger)
+		return step.operationManager.RetryOperationWithoutFail(operation, step.Name(), "Kyma resource still exists", 15*time.Second, step.kymaResourceDeletionTimeout, logger)
```
piotrmiskiewicz, line 65: let's decrease the interval when the timeout is so short. Instead of having 15 sec let's have, for example 5 sec

piotrmiskiewicz (approved):

PR #703, opened 2024-04-23
Timeout for k8s operations

jaroslaw-pieszka:
**Description**

We experienced that updater goroutine once "hung" and no updates were executed till the restart.
To avoid we added a timeout, if timeout is reached, item will be requeued.

- Context with timeout used when listing kymas and when updating kymas

**Related issue(s)**

### DIFF
--- a/internal/kymacustomresource/kyma_cr_updater.go
+++ b/internal/kymacustomresource/kyma_cr_updater.go
@@ -17,6 +17,7 @@ const (
 	namespace               = "kcp-system"
 	subaccountIdLabelKey    = "kyma-project.io/subaccount-id"
 	subaccountIdLabelFormat = "kyma-project.io/subaccount-id=%s"
+	k8sRequestInterval      = 5 * time.Second
 )
 
 type Updater struct {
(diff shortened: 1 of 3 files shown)


internal/kymacustomresource/kyma_cr_updater.go
```diff
@@ -17,6 +17,7 @@ const (
 	namespace               = "kcp-system"
 	subaccountIdLabelKey    = "kyma-project.io/subaccount-id"
 	subaccountIdLabelFormat = "kyma-project.io/subaccount-id=%s"
+	k8sRequestInterval      = 5 * time.Second
```
ralikio, line 20: Can you create a test that sets shorter timeout and asserts that function execution fails with timeout?

ralikio (approved):

internal/kymacustomresource/kyma_cr_updater.go
jaroslaw-pieszka: Too much hassle.

ralikio (approved):

PR #692, opened 2024-04-22
Subaccount sync configuration fixes

jaroslaw-pieszka:
**Description**

Changes proposed in this pull request:

- ...
- ...
- ...

**Related issue(s)**

PR #710, opened 2024-04-25
Configure scraping for sub sync

jaroslaw-pieszka:
**Description**

To make metrics exposed by `subaccount-sync` app visible in Plutono we need to add some resources.

- Port added to deployment
- Service added
- VMServiceScrape added
- PeerAuthentication added

**Related issue(s)**

#692


ralikio (approved):

ralikio (changes requested):
Can we used `app.kubernetes.io/name: subaccount-sync` instead of `app: subaccount-sync`?

PR #724, opened 2024-04-30
operation result metrics test

ukff:
- add unit tests for kcp_keb_v2_operation_result


cmd/broker/broker_suite_test.go
```diff
@@ -453,6 +461,25 @@ func (s *BrokerSuiteTest) WaitForOperationState(operationID string, state domain
...
+	var op *internal.Operation
+	err := s.poller.Invoke(func() (done bool, err error) {
+		op, err = s.db.Operations().GetOperationByID(operationID)
+		if err != nil {
+			return false, nil
+		}
+		return true, nil
+	})
+
+	if dberr.IsNotFound(err) {
```
jaroslaw-pieszka, line 474: the same result in both cases.

jaroslaw-pieszka, line 474: So in case of any error we return with `nil`. Why so complex logic?

jaroslaw-pieszka: Don't you think that this would suffice:
`
if err != nil {
   return nil
}
`

cmd/broker/broker_suite_test.go
```diff
@@ -453,6 +461,25 @@ func (s *BrokerSuiteTest) WaitForOperationState(operationID string, state domain
 	assert.NoError(s.t, err, "timeout waiting for the operation expected state %s != %s. The existing operation %+v", state, op.State, op)
 }
 
+func (s *BrokerSuiteTest) GetOperation(operationID string) *internal.Operation {
+	var op *internal.Operation
+	err := s.poller.Invoke(func() (done bool, err error) {
+		op, err = s.db.Operations().GetOperationByID(operationID)
+		if err != nil {
```
jaroslaw-pieszka, line 467: `return err != nil, nil`

jaroslaw-pieszka (approved):

jaroslaw-pieszka (dismissed):

jaroslaw-pieszka (approved):

PR #734, opened 2024-05-07
Lager Logger for Filtering Specific Messages

ralikio:
**Description**

Current configuration of KEBs broker-api uses errors as a way to return [HTTP status code](https://github.com/kyma-project/kyma-environment-broker/blob/6e54bd5fd0494c3c6d7f5be0f4fac4e75a699cf4/internal/broker/instance_get.go#L56). Such errors are logged with `ERROR` level but they do not interrupt normal application behaviour nor require additional intervention from dev side. The following PR introduces possibility to filter such logs by their message. Errors that are filtered out are redirected to stdout with level changed to `INFO`.

- Filtering messages so that unnecessary ERRORs are not logged in broker-api layer.

**Related issue(s)**

### DIFF
--- a/resources/keb/files/subaccount-sync-dashboard.json
+++ b/resources/keb/files/subaccount-sync-dashboard.json
@@ -327,67 +327,6 @@
       "type": "gauge"
     },
     {
-      "datasource": null,
-      "description": "Calls to events and accounts endpoints",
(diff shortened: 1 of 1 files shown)


ralikio:
Decided not to filter such logs in the code.

PR #746, opened 2024-05-09
Debug log for tracking state deletion

jaroslaw-pieszka:
**Description**

Log would be handy to track state deletions from persistent storage

- Debug log

**Related issue(s)**


szwedm (approved):
/lgtm

Issue #793, opened 2024-06-03
Support new sap-converged-cloud plan region: ap-au-1


PK85:
**AC**
- [x] 0) enhance sap-convereged-cloud plan witn possibility to display given ccee regions for given BTP regions. That means KEB must support mapping between a given BTP region e.g. "cf-eu21" and a list of CCEE regions per Landscape. Configuration must be applied here: https://github.tools.sap/kyma/management-plane-config/tree/main/argoenv/keb for all landscapes. https://github.com/kyma-project/kyma-environment-broker/pull/846
  - [x]  0.1) (converted to additional point by @ralikio) Prepare configuration syntax and define behaviour (with corner cases too) present to PO and team - https://github.com/kyma-project/kyma-environment-broker/pull/812, https://github.com/kyma-project/kyma-environment-broker/pull/818, https://github.com/kyma-project/kyma-environment-broker/pull/820
- [x] 1.1) For BTP region cf-ap10 the sap-converged-cloud plan catalog must return ap-au-1 region only:
```
"region": {
	"enum": [
     		"ap-au-1"
	],
	"minLength": 1,
	"type": "string"
}
```
- [x] 1.2) For BTP region cf-eu20 the sap-converged-cloud plan catalog must return eu-de-1 region only. This is already done, but has old implementation. Must be adjusted.
```
"region": {
	"enum": [
     		"eu-de-1"
	],
	"minLength": 1,
	"type": "string"
}
```			
- [ ] 2) Check if /catalog and /createInstance, /updateInstance validates region value - https://github.com/kyma-project/kyma-environment-broker/pull/886
- [x] 3) Adjust documentation - https://github.com/kyma-project/kyma-environment-broker/pull/859
- [x] 4) Documentation update:
   - [x] Prepare an [internal release note](https://help.sap.com/whats-new/f0d108472a9347d99053658d05ae23e2?Component=Kyma%20Runtime&locale=en-US&state=DRAFT&Valid_as_Of=2024-06-19%3A2024-06-19&q=AP)
   - [x] Update the internal Help Portal documentation:
      - [x] [Regions for the Kyma Environment](https://help.sap.com/docs/btp/sap-business-technology-platform-internal/regions-for-kyma-environment?locale=en-US&state=DRAFT&version=Internal) 
[...]

PR #812, opened 2024-06-06
POC Test for Paring Mappings with Envconfig

ralikio:
**Description**

Testing enconfig to see if this is suitable for parsing mappings from https://github.com/kyma-project/kyma-environment-broker/issues/793.

- Test for checking mapping cases with envconfig.
**Related issue(s)**


ralikio:
Decided to go for the same configuration storage as it is done in eu access whitelists.

Issue #693, opened 2024-04-22
Deploy `subaccount-sync` [betaEnabled] to higher landscapes


jaroslaw-pieszka:
`subaccoount-sync` is deployed to DEV.

Deploy it to higher landscapes in controlled way using dry-run mode (updateResources setting) to rollout changes gradually.

AC:
- [x] on STAGE in dry-run mode, monitoring and alerts configured
- [x] on STAGE in fully operational mode
- [x] on PROD in dry-run mode, monitoring and alerts configured
- [x] on PROD in fully operational mode

PK85:
Blocked, waiting for Plutono dashboards

jaroslaw-pieszka:
Currently `subaccount-sync` is deployed on DEV only, monitoring and alerts are configured for DEV.

jaroslaw-pieszka:
Enabling deployment on STAGE and PROD: https://github.tools.sap/kyma/management-plane-config/pull/5307

jaroslaw-pieszka:
Reducing rate of calls for events: https://github.tools.sap/kyma/management-plane-config/pull/5310

On STAGE and PROD we experienced occasionally error status 429 when calling `events` endpoint, so to avoid hitting the limit we reduced  maximal number of call to 3 per 2s.

jaroslaw-pieszka:
https://github.tools.sap/kyma/backlog/issues/5659
is blocking deployment of  https://github.com/kyma-project/kyma-environment-broker/pull/814

PR #817, opened 2024-06-10
Configurable strategy for runtimeID to subaccountID mapping

jaroslaw-pieszka:
**Description**

We fetch runtimeID and subaccountID label from Kyma CR. If subaccountID label is not present, we use DB (`instances` table) to get subaccountID.
Introduced flag allows to fetch subaccountID from DB always, disregarding label value.

- Configuration flag (defaulting to false)
- logic based on flag value fetching subaccountID either always from DB or only in case this label is not present in Kyma CR 

**Related issue(s)**

 #693


ralikio (changes requested):
Missing environment variable setup.

ralikio (approved):

Issue #905, opened 2024-07-03
KIM Integration - Phase 1 point 0: Prepare Structures


ralikio:
Lay ground work for https://github.com/kyma-project/kyma-environment-broker/issues/791. Create structures that prepares work for implementing Runtime CR. Additionally, during analysis of https://github.com/kyma-project/kyma-environment-broker/issues/791 we discovered unnecessary indirect dependencies that are not used in our code.

**Acceptance Criteria** 

- [x] 1) Import structures - later on we can mitigate dependency
- [x] 2) Prepare test printing out yaml representation of created object 
- [x] 3) Fill structure with appropriate data, write down what we data we are missing now

jaroslaw-pieszka:
Discussion with @piotrmiskiewicz concerning setting KymaResourceName:

- we need additional step which is called after `GenerateRuntimeIDStep` - when RuntimeID should have been established
- this step will calculate and store in operations `KymaResourceName`
- this step will defensively check if RuntimeID is set on entry, returning error if not.

jaroslaw-pieszka:
Loose ends:
- [ ] we pass `cfg.Provisioner` to `CreateRuntimeResource` step. We need to handle this differently when Provisioner is no longer used, at least rename it, possibly merge with cfg.KimConfig as Parameter Object.

jaroslaw-pieszka:
Next steps (short-term plans):
- [ ] complete logic creating shoot data for all plans
- [ ] test it mostly on step level
- [ ] create step checking Runtime CR status 
- [ ] tackle deprovisioning

ralikio:
TODO:
- [ ] tests for `internal/provider/gcp.go` to be done after merging of #942

jaroslaw-pieszka:
Next step
- [ ] Networking for Preview, Azure, GCP, AWS
- [ ] Networking for Trial (AWS, GCP, Azure), Sap Converged Cloud, Free, Azure Lite

Issue #791, opened 2024-06-03
KIM migration - KEB: Migrate from provisioner to Kyma Infrastructure Manager


PK85:
### **Phase 1**: (driving system Provisioner for creation, Runtime CR spec is not passed to gardener, Runtime CR status is already present)
- [x] 0) Import (if no dependencies added)/copy CRD structures from KIM - https://github.com/kyma-project/kyma-environment-broker/issues/905 - for the time being we decided that we will handle this later on.
- [x] 1)  use preview plan to use provisioner and KIM - currently configurable for any plan of preference
- [x] 2) create [Runtime CR](https://github.com/kyma-project/infrastructure-manager/blob/5053ed10c903d3cee01b233e08e436d9ef35b7ad/api/v1/runtime_types.go#L79) ~and create [GardenerCluster](https://github.com/kyma-project/infrastructure-manager/blob/5053ed10c903d3cee01b233e08e436d9ef35b7ad/api/v1/gardenercluster_types.go#L34)~ (for later: do not create GardenerCluster CR, just watch a secret by convention name)
  - [x] 2.1 Do not create Gardener Cluster, KIM Team decided to manage it on their own: https://github.com/kyma-project/infrastructure-manager/issues/282
- [x] 3) support update operation which call Provisioner and patch Runtime CR #1068 
- [x] 4) support deprovisioning operation which call Provisioner and delete Runtime CR
- [x] 5) support suspension/expiration operation which call Provisioner and delete Runtime CR #1069 
- [x] 6) ~~support upgrade cluster operation which call Provisioner~~ and delete Runtime CR (this one we need to double check, cause with Runtime CR SRE will upgrade cluster just by modifying Runtime CR directly) #1072  Upgrade would be handled by SRE
- [x] 7) check status from Runtime CR, maybe try to error shoot creation and check a status syntax
- [x] 8) update kcp CLI and `runtimes` endpoint to support instances not known to provisioner (KIM driven) and to support fetching Runtime CR spec. #1070 
- [x] 9) adjust SKR tests to use Runtime config or Gardener config returned by kcp CLI call (extension in #1070 are needed) #1077 
- [x] 10) refactor to reduce technological debt.

**Expected result from phase 1**:
- we figured out all not working scenarios, sytax etc
- provisioning and update works, and deprovisioning
- KIM controller works as expected and status from provisioner and KIM should be generally the same (READY, ERROR ), shoot from provisioner and shoot CR from KIM are the same
[...]

PR #937, opened 2024-07-17
Setting valid Kyma Resource Name in a new step

jaroslaw-pieszka:
**Description**

- Step calculating Kyma Resource name using `step.KymaName` method as a single point of truth, the value is used to update the operation
- removal of no longer needed CreateKymaForOwnCluster step

**Related issue(s)**

#791 
#905


internal/process/provisioning/create_kyma_name_step.go
```diff
@@ -0,0 +1,45 @@
...
+		return s.operationManager.OperationFailed(operation, fmt.Sprintf("operation has reached the time limit: %s", CreateRuntimeTimeout), nil, log)
+	}
+
+	if operation.RuntimeID == "" {
+		return s.operationManager.OperationFailed(operation, fmt.Sprint("RuntimeID not set, cannot create Kyma name"), nil, log)
+	}
+
+	// could be simplified but this provides single source of truth for Kyma name
+	operation.KymaResourceName = ""
+	operation.KymaResourceName = steps.KymaName(operation)
```
ralikio, line 40: Is that duplicated assignment required?

cmd/broker/provisioning.go
```diff
@@ -84,6 +84,13 @@ func NewProvisioningProcessingQueue(ctx context.Context, provisionManager *proce
 			stage:     createRuntimeStageName,
 			step:      provisioning.NewGenerateRuntimeIDStep(db.Operations(), db.RuntimeStates(), db.Instances()),
 		},
+		// postcondition: operation.RuntimeID is set
+		{
+			condition: provisioning.SkipForOwnClusterPlan,
+			stage:     createRuntimeStageName,
+			step:      provisioning.NewCreateKymaNameStep(db.Operations()),
```
ralikio, line 91: Is this steps going to do anything else? For now it feels like it is just doing one assignment? I do not think that it is necessary to create a new step for such assignment. Maybe we could put that in initialisation step or somewhere else?

internal/process/steps/lifecycle_manager.go
```diff
@@ -54,6 +54,10 @@ func KymaName(operation internal.Operation) string {
 	return strings.ToLower(operation.RuntimeID)
 }
 
+func KymaRuntimeResourceName(operation internal.Operation) string {
```
ralikio, line 57: Where is this used?

cmd/broker/provisioning.go
jaroslaw-pieszka: This is how we decided to approach after discussion.

internal/process/steps/lifecycle_manager.go
jaroslaw-pieszka: We decided to create single point of truth for Runtime Resource as well. I will add invocation in next commit.

internal/process/provisioning/create_kyma_name_step.go
ralikio: I do not understand why line 40 is required. I guess it is just to force KymaResourceName to be empty so that `steps.KymaName` for sure sets it to `RuntimeId`. Is that correct? It seems confusing because we need to get into details of invoked method when analysing what the empty assignment is all about.

internal/process/steps/lifecycle_manager.go
jaroslaw-pieszka: Initially I wanted to add invocation in separate PR, but it is more consistent if I add it here.

cmd/broker/provisioning.go
jaroslaw-pieszka: See https://github.com/kyma-project/kyma-environment-broker/issues/905#issuecomment-2230587326

jaroslaw-pieszka: In my opinion this approach is justified, if you have doubts we can arrange a meeting tomorrow.
Some pros:
- simple logic, 
- postcondition guaranteed,
- easy to manage in the process of migration from Provisioner to KIM - we will remove some steps gradually
- easy to extend
- easy to test

Currently we have a lot of quite complex steps, and it is hard to grasp what is set where, could we remove step, could we add a new one, and where to put it, to have all required data i to provide all required data for subsequent steps.

internal/process/provisioning/create_kyma_name_step.go
jaroslaw-pieszka: I have the same feeling. It seems awkward. I need to reconsider this with @piotrmiskiewicz if this is what we wanted and if we have to pay such a price for this `single point of truth`.
I would prefer to change steps.KymaName() semantics and avoid `emptying` the operation.KymaResourceName.
But this function is already used exactly in this way.

jaroslaw-pieszka: I tried to eat a cake and have a cake. Please take a look.

cmd/broker/provisioning.go
piotrmiskiewicz: the question is where we should put this assignment. What are hard requirements? Such assignment (logic, which calculates the Kyma resoruce name) must:
1. be executed after runtime ID gathering/generation (after interaction with Provisioner)
2. be before Runtime CR creation which must be before ApplyKyma step (currently the logic is in the apply kyma step)

@ralikio  - I know this logic is very simple, but we can't put it into initialisation step or something like this (it is too early).
We can't put it into a provisioner interaction step (we gather runtime ID there), because this step will be removed soon.

internal/process/provisioning/create_kyma_name_step.go
```diff
@@ -0,0 +1,44 @@
...
+	if time.Since(operation.UpdatedAt) > CreateRuntimeTimeout {
+		log.Infof("operation has reached the time limit: updated operation time: %s", operation.UpdatedAt)
+		return s.operationManager.OperationFailed(operation, fmt.Sprintf("operation has reached the time limit: %s", CreateRuntimeTimeout), nil, log)
+	}
+
+	if operation.RuntimeID == "" {
+		return s.operationManager.OperationFailed(operation, fmt.Sprint("RuntimeID not set, cannot create Kyma name"), nil, log)
+	}
+
+	operation.KymaResourceName = steps.CreateKymaNameFromOperation(operation)
```
piotrmiskiewicz, line 39: this line should be inside the inline function in the UpdateOperation

internal/process/provisioning/create_kyma_name_step.go
```diff
@@ -0,0 +1,44 @@
...
+
+	return s.operationManager.UpdateOperation(operation, func(op *internal.Operation) {
+		op.RuntimeID = operation.RuntimeID
```
piotrmiskiewicz, line 42: we do not need to update RuntimeID

internal/process/provisioning/create_kyma_name_step.go
```diff
@@ -0,0 +1,44 @@
...
+		operationManager: process.NewOperationManager(os),
+	}
+}
+
+func (s *CreateKymaNameStep) Name() string {
+	return "Create_Kyma_Name"
+}
+
+func (s *CreateKymaNameStep) Run(operation internal.Operation, log logrus.FieldLogger) (internal.Operation, time.Duration, error) {
+	if time.Since(operation.UpdatedAt) > CreateRuntimeTimeout {
```
piotrmiskiewicz, line 30: In my opinion, we do not need to check any timeout

internal/process/provisioning/create_runtime_resource_step.go
```diff
@@ -58,7 +58,8 @@ func (s *CreateRuntimeResourceStep) Run(operation internal.Operation, log logrus
 		return operation, 0, nil
 	}
 
-	kymaResourceName, kymaResourceNamespace := getKymaNames(operation)
+	kymaResourceName := steps.KymaRuntimeResourceName(operation)
```
piotrmiskiewicz, line 61: once, we set the KymaName in the previous step, we can just rely on an operation field

internal/process/provisioning/create_kyma_name_step.go
jaroslaw-pieszka: Removed in the next commit.

internal/process/provisioning/create_runtime_resource_step.go
jaroslaw-pieszka: Changed in the next commit.

internal/process/provisioning/create_kyma_name_step.go
jaroslaw-pieszka: In the next commit

internal/process/provisioning/create_kyma_name_step.go
jaroslaw-pieszka: in the next commit

cmd/broker/provisioning.go
```diff
@@ -84,6 +84,13 @@ func NewProvisioningProcessingQueue(ctx context.Context, provisionManager *proce
 			stage:     createRuntimeStageName,
 			step:      provisioning.NewGenerateRuntimeIDStep(db.Operations(), db.RuntimeStates(), db.Instances()),
 		},
+		// postcondition: operation.RuntimeID is set
+		{
+			condition: provisioning.SkipForOwnClusterPlan,
```
piotrmiskiewicz, line 89: we do not need this condition, it should wokr also for own cluster

jaroslaw-pieszka: In the next commit.

jaroslaw-pieszka: But then test for OwnCluster fail.
Need to investigate.

jaroslaw-pieszka: Removed no longer needed step.

piotrmiskiewicz (approved):

piotrmiskiewicz (approved):

PR #976, opened 2024-08-01
KIM Integration - AWS Freemium and Azure Lite

jaroslaw-pieszka:
**Description**

- Shoot data for Azure Lite
- Shoot data for AWS Freemium 

**Related issue(s)**

### DIFF
--- a/internal/process/provisioning/create_runtime_resource_step.go
+++ b/internal/process/provisioning/create_runtime_resource_step.go
@@ -257,7 +257,7 @@ func (s *CreateRuntimeResourceStep) providerValues(operation *internal.Operation
 		case internal.AWS:
 			p = &provider.AWSTrialInputProvider{
 				PlatformRegionMapping:  s.trialPlatformRegionMapping,
-				UseSmallerMachineTypes: s.useSmallerMachinesForTrials,
+				UseSmallerMachineTypes: s.useSmallerMachineTypes,
 				ProvisioningParameters: operation.ProvisioningParameters,
 			}
 		case internal.GCP:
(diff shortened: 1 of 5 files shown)


internal/process/provisioning/create_runtime_resource_step.go
```diff
@@ -228,6 +228,24 @@ func (s *CreateRuntimeResourceStep) providerValues(operation *internal.Operation
 			MultiZone:              s.config.MultiZoneCluster,
 			ProvisioningParameters: operation.ProvisioningParameters,
 		}
+	case broker.FreemiumPlanID:
+		var freemiumProvider internal.CloudProvider
+		if operation.ProvisioningParameters.Parameters.Provider == nil {
+			//TODO - default provider?
+			return provider.Values{}, fmt.Errorf("provider not definded for freemium plan")
+		} else {
```
ralikio, line 236: Can we move it outside of else? Because of return it is not necessary.

jaroslaw-pieszka: `else` removed in the next push.

internal/provider/aws.go
```diff
@@ -87,3 +90,29 @@ func (p *AWSTrialInputProvider) region() string {
...
 }
+
+func (p *AWSFreemiumInputProvider) Provide() Values {
+	machineType := DefaultOldAWSTrialMachineType
+	if p.UseSmallerMachineTypes {
+		machineType = DefaultAWSMachineType
+	}
+	region := p.region()
+
+	return Values{
```
ralikio, line 101: There is a lot of inputs set here: https://github.com/kyma-project/kyma-environment-broker/blob/b4a26925fa1ae24f1f063e6d12e92146b6aed064/internal/provider/azure_provider.go#L197-L198. Is that the correct place that should server as a template of what to sent to RunimeResource?

internal/provider/azure.go
```diff
@@ -93,3 +96,36 @@ func (p *AzureTrialInputProvider) region() string {
 	}
 	return DefaultAzureRegion
 }
+
+func (p *AzureLiteInputProvider) Provide() Values {
```
ralikio, line 100: The same question as defined here: https://github.com/kyma-project/kyma-environment-broker/pull/976/files#r1700033908

ralikio (approved):

Issue #831, opened 2024-06-11
KEB: Add new/adjust GitAction job that tests if we can install chart by helm


PK85:
**Reason**

Already happen twice that KEB release was green, and then in the environment some pods were failing because of missing configuration.

**AC**

1) Extends this guy https://github.com/kyma-project/kyma-environment-broker/blob/main/.github/workflows/run-keb-chart-tests-reusable.yaml to do helm install too.
2) Use initialization.yaml from Maper (check with him) 
3) Check if all deployments, jobs are created and deployed without issues.

jaroslaw-pieszka:
Ad 1. If we want to use `helm install` instead of `helm template . | kubectl apply -f -` we need to create all resources needed to deploy keb successfully. It's doable but we create environment more and more different than DEV landscape.

MarekMichali:
We decided to switch to image-builder This will allow us to create images on demand without being tied to tag creation. This will allow us to create a cleaner testing workflow with the helm install.

PK85:
blocked waiting for Marek back from vacation

PR #979, opened 2024-08-02
Run helm install test in create release without prow workflow

MarekMichali:
**Description**

This PR prepares the new release workflow that uses the GitHub Actions image-build to use helm install. While we are still using the old release workflow with prow, the `run-keb-chart-install-tests-reusable.yaml` file has been temporarily duplicated to not break the old workflow.

- Add postgres config for tests
- Add temporary `run-keb-chart-install-tests-reusable.yaml` workflow
- Run helm install in release without prow

**Related issue(s)**

See #831

### DIFF
--- a/.github/workflows/create-release-without-prow.yaml
+++ b/.github/workflows/create-release-without-prow.yaml
@@ -63,6 +63,7 @@ jobs:
       release: "true"
   
   build-keb-image:
+    needs: [validate-release]
     uses: kyma-project/test-infra/.github/workflows/image-builder.yml@main
     with: 
       name:  kyma-environment-broker/ga
@@ -71,6 +72,7 @@ jobs:
       tags: ${{ inputs.name }}
 
   build-archiver-image:
+    needs: [validate-release]
     uses: kyma-project/test-infra/.github/workflows/image-builder.yml@main
     with: 
       name: kyma-environment-archiver-job/ga
@@ -80,6 +82,7 @@ jobs:
       tags: ${{ inputs.name }}
 
   build-environments-cleanup-image:
+    needs: [validate-release]
     uses: kyma-project/test-infra/.github/workflows/image-builder.yml@main
     with: 
       name: kyma-environments-cleanup-job/ga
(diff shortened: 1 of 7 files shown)


.github/workflows/create-release-without-prow.yaml
```diff
@@ -133,10 +141,10 @@ jobs:
       build-args: BIN=subaccount-sync
       tags: ${{ inputs.name }}
 
-  run-keb-chart-tests:
+  run-keb-chart-install-tests:
     name: Validate KEB chart 
     needs: [build-keb-image, build-archiver-image, build-environments-cleanup-image, build-deprovision-retrigger-image, build-expirator-image, build-runtime-reconciler-image, build-subaccount-cleanup-image, build-subaccount-sync-image]
-    uses: "./.github/workflows/run-keb-chart-tests-reusable.yaml"
+    uses: "./.github/workflows/run-keb-chart-install-tests-reusable.yaml"
```
ralikio, line 147: Is `run-keb-chart-tests-reusable.yaml` still used somewhere?

MarekMichali: In the current release workflow where we are using prow.

MarekMichali: And in tests on PR's.

.github/workflows/run-keb-chart-install-tests-reusable.yaml
```diff
@@ -0,0 +1,184 @@
...
+      with:
+        files: |
+          .github/**
+          resources/**
+          scripts/**
+          Makefile
+          **/*.sh
+
+    - name: Prepare chart for next KEB version
+      if: inputs.release == 'true'
```
ralikio, line 39: do we have to use comparison in this iff statement?

MarekMichali: Yes, this is a string.

ralikio (approved):

.github/workflows/create-release-without-prow.yaml
ralikio: This is just a temporary for created for test convenience.

.github/workflows/run-keb-chart-install-tests-reusable.yaml
ralikio: We have agreed to check if boolean works later on.

PR #1000, opened 2024-08-07
KIM Integration - Testing networking for non-default parameters

jaroslaw-pieszka:
**Description**

- Test case for non-default parameters

**Related issue(s)**

#791 
#905


internal/process/provisioning/create_runtime_resource_step_test.go
```diff
@@ -378,6 +378,62 @@ func TestCreateRuntimeResourceStep_Defaults_AWS_SingleZone_DryRun_ActualCreation
...
+	// given
+	log := logrus.New()
+	memoryStorage := storage.NewMemoryStorage()
+
+	err := imv1.AddToScheme(scheme.Scheme)
+
+	instance, operation := fixInstanceAndOperation(broker.AWSPlanID, "eu-west-2", "platform-region")
+	operation.ProvisioningParameters.Parameters.Networking = &internal.NetworkingDTO{
+		NodesCidr:    "192.168.48.0/20",
+		PodsCidr:     ptr.String("192.168.48.0/20"),
```
piotrmiskiewicz, line 391: you have the same values for pods and nodes, I think we should have 3 different ranges

jaroslaw-pieszka: My bad. Fix in the next commit.

piotrmiskiewicz (approved):

PR #1026, opened 2024-08-09
KIM Integration - Volume support and worker group name

jaroslaw-pieszka:
**Description**

- support for volume size, volume type default values
- support of worker group name (hard-wired)
- volume size populated from parameters (if present)

**Related issue(s)**

#791 
#905


szwedm (commented):
What do you think about adding consts for all repeating values? I would like to know more about the origin of values, let's have a short call.

jaroslaw-pieszka:
> What do you think about adding consts for all repeating values? I would like to know more about the origin of values, let's have a short call.

It is supposed to be next step in the refactoring. We are working on test in small increments, still a lot to cover and improve.

szwedm (approved):
/lgtm

Issue #1021, opened 2024-08-09
KCP CLI - List of all deprovisioned instances should not be allowed


KsaweryZietara:
**AC**
- limit result to up to 1000 last deprovisioned ones regardless of the input filter 
https://github.com/kyma-project/kyma-environment-broker/pull/1022
https://github.com/kyma-project/kyma-environment-broker/pull/1028, [#23](https://github.tools.sap/kyma/kcp-cli/pull/23)
- document in the kcp rt --help [#21](https://github.tools.sap/kyma/kcp-cli/pull/21)

PR #1028, opened 2024-08-12
Limit deprovisioned instances

KsaweryZietara:
**Description**

- revert changes from https://github.com/kyma-project/kyma-environment-broker/pull/1022,
- limit deprovisioned instances in client.

**Related issue(s)**
See also https://github.com/kyma-project/kyma-environment-broker/issues/1021

### DIFF
--- a/common/runtime/client.go
+++ b/common/runtime/client.go
@@ -12,7 +12,10 @@ import (
 	"github.com/kyma-project/kyma-environment-broker/common/pagination"
 )
 
-const defaultPageSize = 100
+const (
+	defaultPageSize = 100
+	RuntimesLimit   = 1000
+)
 
 // Client is the interface to interact with the KEB /runtimes API as an HTTP client using OIDC ID token in JWT format.
 type Client interface {
@@ -41,10 +44,18 @@ func (c *client) ListRuntimes(params ListParameters) (RuntimesPage, error) {
 	runtimes := RuntimesPage{}
 	getAll := false
 	fetchedAll := false
+	limitRuntimes := false
 	if params.Page == 0 || params.PageSize == 0 {
 		getAll = true
 		params.Page = 1
 		params.PageSize = defaultPageSize
+	}
+
+	for _, state := range params.States {
+		if state == StateDeprovisioned {
+			limitRuntimes = true
+			break
+		}
 	}
 
 	for !fetchedAll {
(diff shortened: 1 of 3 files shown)


common/runtime/client.go
```diff
@@ -12,7 +12,10 @@ import (
 	"github.com/kyma-project/kyma-environment-broker/common/pagination"
 )
 
-const defaultPageSize = 100
+const (
+	defaultPageSize = 100
+	RuntimesLimit   = 1000
```
jaroslaw-pieszka, line 17: So we leave it hard-wired, no configuration possible?

common/runtime/client.go
```diff
@@ -91,6 +102,12 @@ func (c *client) ListRuntimes(params ListParameters) (RuntimesPage, error) {
 		} else {
 			fetchedAll = true
 		}
+
+		if limitRuntimes && runtimes.Count >= RuntimesLimit {
+			runtimes.Data = runtimes.Data[:RuntimesLimit]
+			runtimes.Count = RuntimesLimit
+			fetchedAll = true
```
jaroslaw-pieszka, line 109: Is this meant just to force break out of the for loop?

common/runtime/client_test.go
```diff
@@ -118,6 +118,32 @@ func TestClient_ListRuntimes(t *testing.T) {
 		assert.Equal(t, 4, rp.TotalCount)
 		assert.Len(t, rp.Data, 4)
 	})
+
+	t.Run("Test deprovisioned runtimes limit", func(t *testing.T) {
+		//given
+		params := ListParameters{
```
jaroslaw-pieszka, line 124: Do we test an execution path where there is no `StateDeprovisioned` in the params, so the limit has no effect (`limitRuntimes` is false).

common/runtime/client.go
KsaweryZietara: I think this parameter will not change in the future, so I would leave it hard-wired. Do you think it should be configurable?

common/runtime/client_test.go
KsaweryZietara: There are two tests without `StateDeprovisioned` in the params: 
- https://github.com/kyma-project/kyma-environment-broker/blob/main/common/runtime/client_test.go#L35
- https://github.com/kyma-project/kyma-environment-broker/blob/main/common/runtime/client_test.go#L97

jaroslaw-pieszka (approved):

PR #1061, opened 2024-08-26
KIM Integration - instances on `runtimes` endpoint decorated with Runtime CR

jaroslaw-pieszka:
**Description**

KEB endpoint `runtimes` is used by `kcp` CLI e.g. to get gardener configuration information (using `--gardener-config` switch), and this information is used quite often to assert in SKR tests. Getting gardener configuration requires communication with provisioner, so in case of KIM driven instances is pointless.
We need the way to base assertions on the spec fields of Runtime CR.

- adding new query parameter `runtime-config`
- handling scenario when with `--gardener-config` flag we process instance not known to provisioner

**Related issue(s)**

#791 
#905

### DIFF
--- a/internal/runtime/handler.go
+++ b/internal/runtime/handler.go
@@ -3,8 +3,9 @@ package runtime
 import (
 	"context"
 	"fmt"
-	"k8s.io/apimachinery/pkg/api/errors"
 	"net/http"
+
+	"k8s.io/apimachinery/pkg/api/errors"
 
 	"github.com/kyma-project/kyma-environment-broker/internal/process/steps"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
--- a/internal/runtime/handler.go
+++ b/internal/runtime/handler.go
@@ -1,8 +1,16 @@
 package runtime
 
 import (
+	"context"
 	"fmt"
+	"k8s.io/apimachinery/pkg/api/errors"
 	"net/http"
+
+	"github.com/kyma-project/kyma-environment-broker/internal/process/steps"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/sirupsen/logrus"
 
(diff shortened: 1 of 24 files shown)


internal/runtime/handler.go
```diff
@@ -205,13 +219,31 @@ func (h *Handler) getRuntimes(w http.ResponseWriter, req *http.Request) {
...
+		if runtimeResourceConfig && dto.RuntimeID != "" && instanceDrivenByKimOnly {
+			runtimeResourceName, runtimeNamespaceName := h.getRuntimeNamesFromLastOperation(dto)
+
+			runtimeResourceObject := &unstructured.Unstructured{}
+			runtimeResourceObject.SetGroupVersionKind(RuntimeResourceGVK())
+			err = h.k8sClient.Get(context.Background(), client.ObjectKey{
+				Namespace: runtimeNamespaceName,
+				Name:      runtimeResourceName,
+			}, runtimeResourceObject)
+			if err != nil {
```
piotrmiskiewicz, line 241: I think in case of "Not Found" error we will have a problem here.

internal/runtime/handler.go
```diff
@@ -205,13 +219,31 @@ func (h *Handler) getRuntimes(w http.ResponseWriter, req *http.Request) {
...
+		instanceDrivenByKimOnly := h.kimConfig.IsDrivenByKimOnly(dto.ServicePlanName)
+
+		err = h.setRuntimeOptionalAttributes(&dto, kymaConfig, clusterConfig, gardenerConfig, instanceDrivenByKimOnly)
 		if err != nil {
 			h.logger.Warn(fmt.Sprintf("unable to set optional attributes: %s", err.Error()))
 			httputil.WriteErrorResponse(w, http.StatusInternalServerError, err)
 			return
 		}
 
+		if runtimeResourceConfig && dto.RuntimeID != "" && instanceDrivenByKimOnly {
```
piotrmiskiewicz, line 232: I think the `instanceDrivenByKimOnly` in not necessary here, I'd like to have a possibility to use kcp cli to get the runtime even if it is driven by Provisioner (in the phase when Provisioner is called and runtime resource is created).

internal/runtime/handler.go
piotrmiskiewicz: If there is no runtime resource, let's do not throw any error, just leave the field empty.

jaroslaw-pieszka: We log Warning and return empty result.

piotrmiskiewicz: handle NotFoundError -> log.Info, other error log as warn

internal/runtime/handler.go
jaroslaw-pieszka: In the next commit/push.

internal/runtime/handler.go
jaroslaw-pieszka: In the next commit/push.

piotrmiskiewicz (dismissed):

piotrmiskiewicz (approved):

PR #1062, opened 2024-08-27
Script for sync EDP and KCP (subaccounts)

piotrmiskiewicz:
**Description**

- script for checking Subaccounts in EDP

**Related issue(s)**


utils/edp-registrator/check_all_sids.sh
```diff
@@ -0,0 +1,13 @@
+
```
jaroslaw-pieszka, line 1:
```suggestion
#!/usr/bin/env bash
```

jaroslaw-pieszka (approved):

PR #1071, opened 2024-08-28
Add VaultStaticSecret support to the keb

ujvaritamas:
**Description**
In order to stop using ArgoCD plugins, we need to replace our current mechanism for deploying secrets to the kcp clusters. The Vault Secrets Operator has been deployed to all 3 landscapes, and is now functional.

Everywhere where argo plugin was used -> the logic is changed to use the VSO.

Changes proposed in this pull request: Ticket: https://github.tools.sap/kyma/backlog/issues/5847

- Vault Secrets Operator used to get values from Vault (instead of the argo plugin)
- dev stage prod mount handled with corresponding global values: 

- unused values (what use the argo vault plugin) deleted: https://github.tools.sap/kyma/management-plane-config/pull/5505

**Related issue(s)**
https://github.tools.sap/kyma/backlog/issues/5847


resources/keb/Chart.yaml
```diff
@@ -3,5 +3,5 @@ appVersion: "1.8.34"
 name: keb
 description:
 description: Kyma Environment Broker Helm chart for Kubernetes
-version: 1.8.34
+version: 1.8.35
```
MarekMichali, line 6: Please revert this change. This value is auto bumped by our release workflow.

ujvaritamas: chart version reverted

MarekMichali (approved):

Issue #935, opened 2024-07-16
Accept update operations on `contexts` update attribute for `free` and `trial` expired Kyma clusters


PK85:
**Description**

We consider it be safe to allow updating expired Kyma instances by ERS system. 

**Reasons**
 
The CIS team has issues with `free` and `trial` expired clusters when some nonuser actions are happening, like "Move Subaccount" action etc

**AC**

- [x] 1) For expired Kyma clusters allow to update only "context" attribute, do nothing, respond with success update http status. Updating input parameters by users must be still rejected with 400 bad request status.
- [x] 2) Add feature flag, enable first on staging and we will pass it to CIS team for testing
- [x] 3) Add e2e test/unit tests that expired Kyma is getting context update with new GA, assert that we do not get 400 if only context is changed, check that we still get 400 if update paframeters provided
- [x] 4) Adjust opensource documentation about that scenario and new behaviour for expired clusters

PR #1074, opened 2024-08-29
dont return error if there was context update to expired instance

ukff:
Description

- dont return error http code if there was context update to expired instance, which in practive mean that if only ers context was updated for expired instance we responde with http success code

**Related issue(s)**
See also https://github.com/kyma-project/kyma-environment-broker/issues/935

### DIFF
--- a/cmd/broker/update_test.go
+++ b/cmd/broker/update_test.go
@@ -2363,8 +2363,8 @@ func TestMultipleUpdateNetworkFilterPersisted(t *testing.T) {
 	assert.Nil(suite.t, instance.Parameters.ErsContext.LicenseType)
 
 	// when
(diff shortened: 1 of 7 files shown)


cmd/broker/update_test.go
```diff
@@ -2401,3 +2401,128 @@ func TestMultipleUpdateNetworkFilterPersisted(t *testing.T) {
 	disabled = true
 	suite.AssertDisabledNetworkFilterRuntimeState(instance.RuntimeID, updateOperation2ID, &disabled)
 }
+
+func TestUpdateOnlyErsContext(t *testing.T) {
```
piotrmiskiewicz, line 2405: small: rename the test like: TestUpdateOnlyErsContextForExpiredInstance or something like this (the "expiration" is mentioned in the name)

piotrmiskiewicz (approved):

Issue #992, opened 2024-08-06
Do not retry deprovisioning if Kyma resource was not deleted


PK85:
**Reason:**

After the step Provisioner is called to completely delete the k8s cluster. There is no reason to repeat the check kyma resoruce step when the luster is gone

**AC:**
 - Check Kyma resource deleted must not record the step as "executedNotCompleted" in case of Kyma resource was not deleted

Outcome:
We do not see deprovisionedIncomplete via kcp CLI with reason that "Kyma resource was not deleted properly"

PR #1127, opened 2024-09-12
Do not retry deprovisioning if Kyma resource was not deleted

MarekMichali:
**Description**

if Kyma resource was not deleted don't add the `Check_Kyma_Resource_Deleted` step as deprovision incomplete to avoid retrying the deprovision.

**Related issue(s)**

See #992

### DIFF
--- a/internal/process/deprovisioning/check_kyma_resource_deleted_step.go
+++ b/internal/process/deprovisioning/check_kyma_resource_deleted_step.go
@@ -62,7 +62,7 @@ func (step *CheckKymaResourceDeletedStep) Run(operation internal.Operation, logg
 		Name:      kymaResourceName,
 	}, kymaUnstructured)
 
-	if !errors.IsNotFound(err) && err != nil {
+	if err != nil && !errors.IsNotFound(err) {
 		logger.Errorf("unable to check Kyma resource existence: %s", err)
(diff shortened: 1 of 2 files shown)


internal/process/deprovisioning/check_kyma_resource_deleted_step.go
```diff
@@ -62,16 +62,15 @@ func (step *CheckKymaResourceDeletedStep) Run(operation internal.Operation, logg
 		Name:      kymaResourceName,
 	}, kymaUnstructured)
 
-	if err == nil {
-		logger.Infof("Kyma resource still exists")
-		return step.operationManager.RetryOperationWithoutFail(operation, step.Name(), "Kyma resource still exists", 5*time.Second, step.kymaResourceDeletionTimeout, logger, nil)
-	}
-
-	if !errors.IsNotFound(err) {
+	if !errors.IsNotFound(err) && err != nil {
```
jaroslaw-pieszka, line 65: How isNotFound handles err = nil
Should not we change the order?
```suggestion
	if err != nil && !errors.IsNotFound(err) {
```

MarekMichali: When err is nil then IsNotFound returns false.

jaroslaw-pieszka (approved):

Issue #940, opened 2024-07-19
Subaccount movement to another GA must update labels


piotrmiskiewicz:
Subaccount movement which is an OSB context update must change labels in Kyma/GardenerCluster/Runtime resources.
AC:
 - when a context update with a GA change occurs, proper GA labels in Kyma, Runtime and GardenerCluster resoruces must bu updated
 - prepare a script which updates all resorces, where GA where changed

PR #1158, opened 2024-09-18
Label CRs with new global account id

ukff:
**Description**

when changing global account for instance, then relabel related CRs with new value

**Related issue(s)**

https://github.com/kyma-project/kyma-environment-broker/issues/940


cmd/broker/main.go
```diff
@@ -236,7 +236,7 @@ func main() {
 	}
 
 	cfg.OrchestrationConfig.KubernetesVersion = cfg.Provisioner.KubernetesVersion
-
+	cfg.Broker.UpdateCustomResouresLabelsOnAccountMove = true //temp
```
jaroslaw-pieszka, line 239: Is this for testing only?

internal/broker/instance_update.go
```diff
@@ -347,20 +356,29 @@ func (b *UpdateEndpoint) processContext(instance *internal.Instance, details dom
 		instance.Parameters.ErsContext.Active = ersContext.Active
 	}
 
+	needUpdateCustomResources := false
 	if b.subaccountMovementEnabled {
```
jaroslaw-pieszka, line 360: Could be simplified
```if b.subaccountMovementEnabled && instance.GlobalAccountID != ersContext.GlobalAccountID && ersContext.GlobalAccountID != "" {```

File: internal/broker/instance_update.go [hunk 3]
```diff
@@ -347,20 +356,29 @@ func (b *UpdateEndpoint) processContext(instance *internal.Instance, details dom
...
 		}
-
 	}
 
 	newInstance, err := b.instanceStorage.Update(*instance)
 	if err != nil {
 		logger.Errorf("processing context updated failed: %s", err.Error())
 		return nil, changed, fmt.Errorf("unable to process the update")
+	} else if b.updateCustomResouresLabelsOnAccountMove && needUpdateCustomResources {
+		// update labels on related CRs, but only if account movement was succefully persistent and kept in database
```
jaroslaw-pieszka, line 375:
```suggestion
		// update labels on related CRs, but only if account movement was succefully persisted and kept in database
```

internal/broker/instance_update.go
```diff
@@ -389,3 +407,57 @@ func (b *UpdateEndpoint) getJsonSchemaValidator(provider internal.CloudProvider,
...
+	}
+
+	kcpK8sClient, err := client.New(kcpK8sConfig, client.Options{})
+	if err != nil {
+		return nil, fmt.Errorf("while getting creating kcp cluster from k8s config: %s", err.Error())
+	}
+	return &kcpK8sClient, nil
+}
+
+func (b *UpdateEndpoint) updateCrLabel(id, crName, newGlobalAccountId string) error {
```
jaroslaw-pieszka, line 435: So we create new client for each call...

internal/broker/instance_update.go
```diff
@@ -389,3 +407,57 @@ func (b *UpdateEndpoint) getJsonSchemaValidator(provider internal.CloudProvider,
...
+	}
+
+	var k8sObject unstructured.Unstructured
+	k8sObject.SetGroupVersionKind(gvk)
+	err = (*kcpK8sClient).Get(context.Background(), types.NamespacedName{Namespace: KymaNamespace, Name: id}, &k8sObject)
+	if err != nil {
+		return fmt.Errorf("while getting k8s object of type %s from kcp cluster %s for instance, due to: %s", crName, id, err.Error())
+	}
+
+	err = k8s.AddOrOverrideMetadata(&k8sObject, k8s.GlobalAccountIdLabel, newGlobalAccountId)
```
jaroslaw-pieszka, line 457: Here you update in-memory data of the object. Where the actual update is?

cmd/broker/main.go
jaroslaw-pieszka: Why don't you test it via configuration file without hard coded temporary values?

internal/broker/instance_update_test.go
```diff
@@ -795,3 +807,84 @@ func TestSubaccountMovement(t *testing.T) {
...
+			err := fakeKcpK8sClient.Delete(context.Background(), &us)
+			require.NoError(t, err)
+			return
+		}
+		err = fakeKcpK8sClient.Create(context.Background(), &us)
+		require.NoError(t, err)
+	}
+
+	f(t, k8s.KymaCr, false)
+	defer f(t, k8s.KymaCr, true)
```
jaroslaw-pieszka, line 885: Will defer be called just before we exit `createFakeCrs` function?

jaroslaw-pieszka (approved):

PR #1222, opened 2024-09-30
Adjust workflow skipping

MarekMichali:
**Description**

During release the kyma-gopher-bot creates a PR with bumps. For example, it changes the values.yaml of the KEB chart. We shouldn't run tests on exactly this PR and waste time, because we do tests in the release workflow. But we can't skip those jobs with path filtering either, because human developers also change those files during their daily work and then the tests need to be run.

A workaround is to add a check to workflows when they are triggered by kyma-gopher-bot and skip them. The `All checks passed` workflow only needs to be green to merge a PR, so in the case of kyma-gopher-bot we can just return a success status immediately without waiting for other jobs. 

Note that if kyma-gopher-bot does more _development_ PR's in the future, this behavior will need to be changed so that all workflow states are checked as in the case of a human developer PR.

- Add more path filtering
- Immediately return success in `All check passed` when triggered by kyma-gopher-bot

**Related issue(s)**

### DIFF
--- a/docs/contributor/04-10-workflows.md
+++ b/docs/contributor/04-10-workflows.md
@@ -49,7 +49,7 @@ This [workflow](/.github/workflows/auto-merge.yaml) enables the auto-merge funct
 
 ## All Cheks Passed Workflow
 
-This [workflow](/.github/workflows/pr-checks.yaml) checks if all jobs, except those excluded in the workflow configuration, have passed.
+This [workflow](/.github/workflows/pr-checks.yaml) checks if all jobs, except those excluded in the workflow configuration, have passed. If the workflow is triggered by a PR where the author is the `kyma-gopher-bot`, the workflow will end immediately with success.
 
 ## Reusable Workflows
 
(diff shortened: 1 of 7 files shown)


docs/contributor/04-10-workflows.md
```diff
@@ -49,7 +49,7 @@ This [workflow](/.github/workflows/auto-merge.yaml) enables the auto-merge funct
 
 ## All Cheks Passed Workflow
 
-This [workflow](/.github/workflows/pr-checks.yaml) checks if all jobs, except those excluded in the workflow configuration, have passed.
+This [workflow](/.github/workflows/pr-checks.yaml) checks if all jobs, except those excluded in the workflow configuration, have passed. If the workflow is triggered by a PR where the author is the `kyma-gopher-bot`, the workflow will end immediately with success.
```
IwonaLanger, line 52:
```suggestion
This [workflow](/.github/workflows/pr-checks.yaml) checks if all jobs, except those excluded in the workflow configuration, have passed. If the workflow is triggered by a PR where the author is the `kyma-gopher-bot`, the workflow ends immediately with success.
```

IwonaLanger (approved):

jaroslaw-pieszka (approved):

Issue #1204, opened 2024-09-26
SB: Create e2e SKR test for all new service bindings capabilities


PK85:
Subtask of the EPIC: https://github.com/kyma-project/kyma-environment-broker/issues/284

**AC**
- [x] 1) one e2e SKR test calling KEB APis
  - [x] 1.1) https://github.com/kyma-project/kyma-environment-broker/issues/1239, [PR#1244](https://github.com/kyma-project/kyma-environment-broker/pull/1244) 
  - [x] 1.2) add GET and DELETE test for TOKEN+REQUEST ~~and GARDENER~~ (they must fail if GET has dummy response, and when DSELETE TOKEN_REQUEST previous kubeconfig must NOT work, ~~and for GARDENER it must still work~~) 
https://github.com/kyma-project/kyma-environment-broker/pull/1249
  - [x] 1.3) test validation for `expiration_seconds` with less than 10 min, and more than 2h. Must fail and check validation message.  https://github.com/kyma-project/kyma-environment-broker/pull/1261
  - [x] 1.4) Create 11 bidings, 11th must fail with proper validation message https://github.com/kyma-project/kyma-environment-broker/pull/1283
- [ ] 2) new type of e2e test (could be written in go, special for Mr Wojciech). 
  - [ ] 2.1) Setup: We have prepared subacout on staging with Provisioing Service instance and binding, 
  - [ ] 2.2) Setup: Kyma entitlement added for example to AWS plan
  - [ ] 2.3) New test, where provisioning API creates Kyma cluster + Service Bindings, we can still use KCP CLI tech users to validate some things, or show why it failed. Use the SB kubeconfig and reach Kyma instance.

MarekMichali:
Part 2 about provisioning api was moved to #1341

PR #1249, opened 2024-10-03
GET and DELETE in Kyma Binding SKR test

KsaweryZietara:
**Description**

- `Kubernetes TokenRequest`: test get binding, delete binding and fetch secret using kubeconfig after binding deletion - it should fail,
- `Gardener admin kubeconfig`: test get binding, delete binding and fetch secret using kubecofnig after binding deletion - it should pass (it is not possible to invalidate admin kubeconfig).

**Related issue(s)**
See also #1204


testing/e2e/skr/skr-binding-test/index.js
```diff
@@ -45,15 +47,48 @@ describe('SKR Binding test', function() {
...
+        expect(err.response.status).equal(404);
+        console.log('Got response:');
+        console.log(err.response.data);
+      } else {
+        throw err;
+      }
+    }
+  });
+
+  it('Try to fetch sap-btp-manager secret using binding from Kubernetes TokenRequest', async function() {
```
ralikio, line 73: Let's change name of it case by including that it is expected to fail, ok?

testing/e2e/skr/skr-binding-test/index.js
```diff
@@ -64,10 +99,38 @@ describe('SKR Binding test', function() {
...
+        expect(err.response.status).equal(404);
+        console.log('Got response:');
+        console.log(err.response.data);
+      } else {
+        throw err;
+      }
+    }
+  });
+
+  it('Try to fetch sap-btp-manager secret using binding from Gardener', async function() {
```
ralikio, line 125: Why is this different then the previous "get and expect to fail" case?

KsaweryZietara: Fetch secret using kubeconfig from Gardener after binding deletion should pass because, it is not possible to invalidate admin kubeconfig.

ralikio (approved):

PR #1261, opened 2024-10-04
Add binding expiration tests to SKR tests

MarekMichali:
**Description**

- Add test for expiration value over max value
- Add test for expiration value below min value
- Improve create binding call to return better errors

**Related issue(s)**

See #1204


szwedm (approved):
/lgtm

PR #1275, opened 2024-10-07
Bindings on runtime endpoint

piotrmiskiewicz:
**Description**

- bindings on runtimes endpoint.
- New flags: "bindings" which adds bindings to the result; "with_bindings" which narrows down the result to SKRs with bindings only

**Related issue(s)**


internal/runtime/handler.go
```diff
@@ -506,6 +519,25 @@ func (h *Handler) getFilters(req *http.Request) dbmodel.InstanceFilter {
 	return filter
 }
 
+func (h *Handler) applyBindings(p *pkg.RuntimeDTO) error {
```
jaroslaw-pieszka, line 522: So we read all bindings for given instance and add it to RuntimeDTO.
Could we consider other verb than `apply`?

common/runtime/model.go
```diff
@@ -121,6 +130,8 @@ const (
 	ExpiredParam         = "expired"
 	GardenerConfigParam  = "gardener_config"
 	RuntimeConfigParam   = "runtime_config"
+	BindingsParam        = "bindings"
+	WithBindingsParam    = "with_bindings" // TODO: find better name
```
jaroslaw-pieszka, line 134: What is the meaning? This field is not used yet.

internal/storage/dbmodel/instance.go
```diff
@@ -35,6 +35,7 @@ type InstanceFilter struct {
 	States                       []InstanceState
 	Expired                      *bool
 	DeletionAttempted            *bool
+	BindingExists                *bool
```
jaroslaw-pieszka, line 38: We can have more than 1 binding per instance? 
So the meaning is `is there at least one binding`, correct?

internal/storage/postsql/read.go
```diff
@@ -882,6 +882,10 @@ func addInstanceFilters(stmt *dbr.SelectStmt, filter dbmodel.InstanceFilter) {
 			stmt.Where("instances.deleted_at = '0001-01-01T00:00:00.000Z'")
 		}
 	}
+
+	if filter.BindingExists != nil && *filter.BindingExists {
+		stmt.Where("instances.instance_id IN (SELECT instance_id FROM bindings)")
```
jaroslaw-pieszka, line 887: For each instance we scan entire bindings table since there is no `where` clause in subquery.

jaroslaw-pieszka:
```
broker=> explain select * from instances where instance_id in (select instance_id from bindings);
                                 QUERY PLAN                                  
-----------------------------------------------------------------------------
 Hash Semi Join  (cost=11.35..130.34 rows=60 width=1019)
   Hash Cond: ((instances.instance_id)::text = (bindings.instance_id)::text)
   ->  Seq Scan on instances  (cost=0.00..118.05 rows=105 width=1019)
   ->  Hash  (cost=10.60..10.60 rows=60 width=516)
         ->  Seq Scan on bindings  (cost=0.00..10.60 rows=60 width=516)
(5 rows)
```

jaroslaw-pieszka:
```sql
select * from instances where exists (select instance_id from bindings where bindings.instance_id=instances.instance_id);
```
This could be more optimal (at least not worse).

internal/storage/dbmodel/instance.go
piotrmiskiewicz: yes

jaroslaw-pieszka (approved):

Issue #1203, opened 2024-09-26
SB: Create Provisioning Service on staging and test Kyma creation and binding management


PK85:
Subtask of the EPIC: https://github.com/kyma-project/kyma-environment-broker/issues/284

**AC**
- [x] user Provisioning API Service to work with kyma instance and service bindings, check if contract is ok and everything works as expected:
    - [x] Test on staging - tested on staging manually (Oct 23: We wait for Zahi for his bugfixing)
    - [x] Test on canary
    - [x] Test on production

KsaweryZietara:
Documentation on how to create Kyma Binding via cis Provisioning API: [#664](https://github.tools.sap/kyma/documentation/pull/664)

Issue #284, opened 2023-12-19
SB: Service Binding for Kyma Service Instance - Epic


pbochynski:
**Description**
Provide a service binding for Kyma Runtime instance. The binding should contain all the information to connect to the runtime (KUBECONFIG or all the information required to build a KUBECONFIG). The binding request can create [TokenRequest](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-request-v1/) and return the token in the binding. 

**Reasons**
Users can create Kyma Runtime instances with API or BTP CLI. But the instance cannot be accessed without user being authenticated (OIDC flow). Therefore any automation flow that involves provisioning Kyma and deploying any workload there cannot be easily implemented. The only way to do that is to use custom OIDC provider, but that option is not easy to set up and use.

**Attachments**
Other way to achieve the automation would be https://github.com/kyma-project/kyma/issues/18305

**AC** Added by PK
- 1) POC -  ✅ done[link](https://github.com/kyma-project/kyma-environment-broker/issues/318)
- 2) Implementation on CIS side -  🏗 in progress 
- 3) Implementation on Kyma side -  🏗 in progress 
  - [x] 3.1) Check with CIS what our endpoints must support (not all OSB API bindings options will be send from CIS to us) @PK85 
  - [ ] 3.2) https://github.com/kyma-project/kyma-environment-broker/issues/1203
  - [x] 3.3) https://github.com/kyma-project/kyma-environment-broker/issues/1197
  - [x] 3.4) https://github.com/kyma-project/kyma-environment-broker/issues/1198
  - [x] 3.5) https://github.com/kyma-project/kyma-environment-broker/issues/1199
  - [x] 3.6) https://github.com/kyma-project/kyma-environment-broker/issues/1202
  - [x] 3.7)  https://github.com/kyma-project/kyma-environment-broker/issues/1196
  - [x] 3.8)  https://github.com/kyma-project/kyma-environment-broker/issues/1204
  - [x] 3.9) https://github.com/kyma-project/kyma-environment-broker/issues/1200
  - [x] 3.10) Decide which plan is bindable (Check with PO)
  - [x] 3.11) https://github.com/kyma-project/kyma-environment-broker/issues/1205
- 4.0) Prepare release:
  - 4.1) RN (Iwona + PK)
  - 4.2) Help portal docs (Iwona + PK)
  - 4.3) deploy stage
  - 4.4) deploy prod

**Links**
[...]

PR #1278, opened 2024-10-08
Add request logging for service bindings

KsaweryZietara:
**Description**

- add request logging for service bindings to check if context contains `user_id`.

**Related issue(s)**
See also #1203, https://github.com/kyma-project/kyma-environment-broker/issues/284#issuecomment-2396292417

### DIFF
--- a/internal/middleware/request_logging.go
+++ b/internal/middleware/request_logging.go
@@ -1,0 +1,42 @@
+package middleware
+
+import (
+	"bytes"
+	"io"
+	"net/http"
+	"strings"
+
+	"github.com/gorilla/mux"
+	"github.com/sirupsen/logrus"
+)
+
+func AddRequestLogging(logs logrus.FieldLogger) mux.MiddlewareFunc {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+			if strings.Contains(req.URL.Path, "service_bindings") {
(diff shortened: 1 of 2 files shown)


internal/middleware/request_logging.go
```diff
@@ -0,0 +1,42 @@
...
+	"bytes"
+	"io"
+	"net/http"
+	"strings"
+
+	"github.com/gorilla/mux"
+	"github.com/sirupsen/logrus"
+)
+
+func AddRequestLogging(logs logrus.FieldLogger) mux.MiddlewareFunc {
```
jaroslaw-pieszka, line 13: We were supposed to gradually migrate logging to slog using logrus only when there is no other option due to legacy dependencies.

internal/middleware/request_logging.go
```diff
@@ -0,0 +1,42 @@
...
+					"method":      req.Method,
+					"url":         req.URL.String(),
+					"query":       req.URL.Query(),
+					"headers":     req.Header,
+					"remote_addr": req.RemoteAddr,
+					"host":        req.Host,
+					"user_agent":  req.UserAgent(),
+				}).Info("Request details")
+
+				if req.Body != nil {
```
jaroslaw-pieszka, line 27: We already logged "Request details" - we can consider logging some message if the body is nil.
It can be inferred, but such explicitly put message would make analysis easier.

jaroslaw-pieszka (approved):

Issue #1198, opened 2024-09-26
SB: Implement Service Binding APIs with DB support for all OSB API operations


PK85:
Subtask of the EPIC: https://github.com/kyma-project/kyma-environment-broker/issues/284

**AC**
- [x] 1) implement db schema for storing and deleting service bindings https://github.com/kyma-project/kyma-environment-broker/pull/1213 
- [x] 2) use db layer in the API endpoints https://github.com/kyma-project/kyma-environment-broker/pull/1262, https://github.com/kyma-project/kyma-environment-broker/pull/1305
- [x] 3) up to 10 SBs allowed, the throw validation error about that fact @MarekMichali https://github.com/kyma-project/kyma-environment-broker/pull/1291
- [x] 4) encrypt/decrypt credentials (kubeconfig) https://github.com/kyma-project/kyma-environment-broker/pull/1213
- [x] 5) if SB expired, then remove this bindings from DB (probably job, idea to store deletionTimestamp when ttl from users comes) done in https://github.com/kyma-project/kyma-environment-broker/issues/1347
- [x] 6) If somethng new for users, update documentation

piotrmiskiewicz:
According to OSB API spec - binding creation must be an idempotent operation. Second, third and next reqeust for the same instance/binding ID (with the same parameters) must return HTTP Ok

PK85:
10.10.2024 - Summary from the meeting about using only token requests from SA from SKR and idempotent behaviour:

Legend:
**BindingID**: means: BindingId - InstanceID pair on KEB side

**Contract between Provisioning Service and KEB bindings:**
- [ ] 1) KEB bindings returns 201 or 200 if everythibng ok, then Provisioning Service persist binding on their side
- [ ] 2) KEB bindings returns 400, 409, 500 if something wrong happen and then  Provisioning Service DO NOT persist binding on their side, just pass error to the user

**KEB bindings implementation details:**
- [ ] 1) we return 201 when BindingID was not existing, and was created with success
- [ ] 2) we return 200 when BindingID WAS already existing, input parameters are the same and was created with success (in_progress,error is not success)
- [ ] 3) we return 400 when invalid input params provided
- [ ] 4) we return 409 when BindingID already exists but input params different
- [ ] 5) we return 500 if KEB itself is not able to finalize binding creation

**KEB implementation about handling biinding related resources, like ServiceAccount, ClusterRole, ClusterRoleBidning**

PR #1291, opened 2024-10-09
Limit bindings to max 10

MarekMichali:
**Description**

- Add limit of max 10 not expired bindings

**Related issue(s)**

See #1198

### DIFF
--- a/internal/broker/bind_create.go
+++ b/internal/broker/bind_create.go
@@ -142,11 +142,9 @@ func (b *BindEndpoint) Bind(ctx context.Context, instanceID, bindingID string, d
 	}
 
 	bindingList, err := b.bindingsStorage.ListByInstanceID(instanceID)
-	switch {
-	case dberr.IsNotFound(err):
-		return domain.Binding{}, apiresponses.ErrInstanceDoesNotExist
-	case err != nil:
-		return domain.Binding{}, apiresponses.NewFailureResponse(fmt.Errorf("failed to get bindings for instance %s", instanceID), http.StatusInternalServerError, fmt.Sprintf("failed to get bindings for instance %s", instanceID))
+	if err != nil {
+		message := fmt.Sprintf("failed to list Kyma bindings: %s", err)
+		return domain.Binding{}, apiresponses.NewFailureResponse(fmt.Errorf(message), http.StatusInternalServerError, message)
 	}
 
 	bindingCount := len(bindingList)
--- a/internal/broker/bind_create.go
+++ b/internal/broker/bind_create.go
@@ -150,10 +150,17 @@ func (b *BindEndpoint) Bind(ctx context.Context, instanceID, bindingID string, d
 	}
 
 	bindingCount := len(bindingList)
-	// dont forget expired bindings, talk with WW
 	if bindingCount >= b.config.MaxBindingsCount {
-		message := fmt.Sprintf("maximum number of bindings reached: %d", bindingCount)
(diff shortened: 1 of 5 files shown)


internal/broker/bind_create.go
```diff
@@ -104,6 +105,28 @@ func (b *BindEndpoint) Bind(ctx context.Context, instanceID, bindingID string, d
 		}
 	}
 
+	bindingList, err := b.bindingsStorage.ListByInstanceID(instanceID)
+	switch {
+	case dberr.IsNotFound(err):
+		return domain.Binding{}, apiresponses.ErrInstanceDoesNotExist
+	case err != nil:
+		return domain.Binding{}, apiresponses.NewFailureResponse(fmt.Errorf("failed to get bindings for instance %s", instanceID), http.StatusInternalServerError, fmt.Sprintf("failed to get bindings for instance %s", instanceID))
```
jaroslaw-pieszka, line 113: According to OSB API we cannot return 404 on the PUT request

internal/broker/bind_create.go
```diff
@@ -104,6 +105,28 @@ func (b *BindEndpoint) Bind(ctx context.Context, instanceID, bindingID string, d
 		}
 	}
 
+	bindingList, err := b.bindingsStorage.ListByInstanceID(instanceID)
+	switch {
+	case dberr.IsNotFound(err):
+		return domain.Binding{}, apiresponses.ErrInstanceDoesNotExist
```
jaroslaw-pieszka, line 111: According to OSB API we cannot return 404 on the PUT request

internal/broker/bind_create_test.go
```diff
@@ -378,6 +392,44 @@ func TestCreateBindingEndpoint(t *testing.T) {
 		assert.Equal(t, firstInstanceSecondBindingDB.Kubeconfig, binding.Credentials.(map[string]interface{})["kubeconfig"])
 		assertClusterAccess(t, response, "secret-to-check-first", binding)
 	})
+
+	t.Run("should return error when maximum number of bindings is reached", func(t *testing.T) {
```
jaroslaw-pieszka, line 396:
```suggestion
	t.Run("should return error when maximum number of bindings is reached before adding a new one", func(t *testing.T) {
```

internal/broker/bind_create.go
MarekMichali: I will change it to 500 as this may be an internal KEB error even when the user provided valid params.

internal/broker/bind_create_test.go
MarekMichali: `should return an error when attempting to add a new bind when the maximum number of bindingshas already been reached` Maybe this is better?

internal/broker/bind_create.go
jaroslaw-pieszka: I am not sure if 500 is the best option. Could we differentiate internal error and not existing instance to return either 400 or 500 accordingly?

MarekMichali: This would require to review the whole bindings code in this PR as we use widely 500. There is a separate subtask for this
https://github.com/kyma-project/kyma-environment-broker/issues/1198#issuecomment-2404880767

jaroslaw-pieszka: We decided to leave 500 as it is and differentiate two cases later on in scope of other ticket.

jaroslaw-pieszka (approved):

jaroslaw-pieszka (approved):

PR #1305, opened 2024-10-10
Initial Bindings Delete API Implementation

ralikio:
**Description**

- injecting bindings storage into bind_delete endpoint,
- added instanceID to the delete storage method to uniquely identify a binding,
- delete method invocation in API endpoint,
- happy path test.

**Related issue(s)**

#284 #1198


piotrmiskiewicz:
I think we need at least one small happy path binding deletion test in `cmd/broker`. It can be a next PR if you like

internal/broker/bind_create_test.go
```diff
@@ -376,11 +378,33 @@ func TestCreateBindingEndpoint(t *testing.T) {
...
+		// when
+		path := fmt.Sprintf("v2/service_instances/%s/service_bindings/%s?accepts_incomplete=false&service_id=%s&plan_id=%s", instanceID, createdBindingID, "123", fixture.PlanId)
+
+		response := CallAPI(httpServer, http.MethodDelete, path, "", t)
+		defer response.Body.Close()
+
+		// then
+		assert.Equal(t, http.StatusOK, response.StatusCode)
+		createdBindingIDDB, err = db.Bindings().Get(instanceID, createdBindingID)
+		assert.Error(t, err)
```
jaroslaw-pieszka, line 401: Is there a way to check if we get 404 not any error?

internal/broker/bind_create_test.go
```diff
@@ -376,11 +378,33 @@ func TestCreateBindingEndpoint(t *testing.T) {
 		binding = unmarshal(t, response)
 		assert.Equal(t, firstInstanceSecondBinding, binding)
 		assert.Equal(t, firstInstanceSecondBindingDB.Kubeconfig, binding.Credentials.(map[string]interface{})["kubeconfig"])
-		assertClusterAccess(t, response, "secret-to-check-first", binding)
+		assertClusterAccess(t, "secret-to-check-first", binding)
+	})
+
+	t.Run("should delete created binding", func(t *testing.T) {
```
jaroslaw-pieszka, line 384: Is there a test case for deleting already deleted resource (idempotency)?

internal/broker/bind_create_test.go
```diff
@@ -376,11 +378,33 @@ func TestCreateBindingEndpoint(t *testing.T) {
...
+		// given
+		instanceID := "1"
+		createdBindingID, createdBinding := createBindingForInstance(instanceID, httpServer, t)
```
jaroslaw-pieszka, line 387: Consider creation of more bindings to check if we delete selectively.
Having delete implemented by `DELETE * from bindings` we would pass this test.

internal/broker/bind_create_test.go
```diff
@@ -376,11 +378,33 @@ func TestCreateBindingEndpoint(t *testing.T) {
...
+		createdBindingIDDB, err := db.Bindings().Get(instanceID, createdBindingID)
+		assert.NoError(t, err)
+		assert.Equal(t, createdBinding.Credentials.(map[string]interface{})["kubeconfig"], createdBindingIDDB.Kubeconfig)
+
+		// when
+		path := fmt.Sprintf("v2/service_instances/%s/service_bindings/%s?accepts_incomplete=false&service_id=%s&plan_id=%s", instanceID, createdBindingID, "123", fixture.PlanId)
```
jaroslaw-pieszka, line 393: Code duplication - consider extraction (preparation of URL and API call).

internal/broker/bind_create_test.go
jaroslaw-pieszka: Why don't we pass `instanceID` and `bindingID` to `createBindingForInstance`?
Instead of creating random bindingID in the function?
We would have more control (using constant or creating random value on the caller level), it would be closer to real usage where the API caller passes both IDs.

ralikio: I am trying to chose the lesser of two evils. I am passing only instanceID since the value is predefined. All tests of bind_create_test.go file rely on two instances that need secrets with Kubeconfig to be created in the current K8S context. I do not want to do this setup in every test case to make the test faster. 

In case of bindingID I do not want to pass it as parameter because `createBindingForInstance` has already too many parameters. Additionally, generating the values make test repeatable because every time we relay on unique values that are not share between the test cases.

internal/broker/bind_create_test.go
ralikio: That is not the case for `DELETE` binding API endpoint. From the spec https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#response-9 we learn that: "
410 Gone | MUST be returned if the Service Binding does not exist."

jaroslaw-pieszka: Agreed to be address in separate PR.

internal/broker/bind_create_test.go
jaroslaw-pieszka: Disagree - three parameters are not too many.

internal/broker/bind_create_test.go
jaroslaw-pieszka: Agreed to be done in separate PR.

jaroslaw-pieszka (approved):

internal/broker/bind_create_test.go
ralikio: I have extracted additional const fields instead of another function creation.

ralikio:
> I think we need at least one small happy path binding deletion test in `cmd/broker`. It can be a next PR if you like

@piotrmiskiewicz I intended `bind_create_test.go` to serve that purpose. However, the placement was incorrect. I have moved the file to broker package as well as switched in memory database to the one created in a container.

jaroslaw-pieszka (approved):

PR #1306, opened 2024-10-11
Add expires_at to bindings table in db

MarekMichali:
**Description**

This change is a result of a meeting where we decided to add the expires_at file to the DB to avoid to calculate this value every time we need it.

- Add expires_at to DB
- Calculate and insert this value to the DB when creating bindings

**Related issue(s)**

See #1198

### DIFF
--- a/internal/broker/bind_create.go
+++ b/internal/broker/bind_create.go
@@ -123,6 +123,7 @@ func (b *BindEndpoint) Bind(ctx context.Context, instanceID, bindingID string, d
 		InstanceID: instanceID,
 
 		CreatedAt: time.Now(),
+		ExpiresAt: time.Now().Add(time.Duration(int64(expirationSeconds)) * time.Second),
 		UpdatedAt: time.Now(),
 
 		ExpirationSeconds: int64(expirationSeconds),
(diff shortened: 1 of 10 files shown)


internal/broker/bind_create.go
```diff
@@ -123,6 +123,7 @@ func (b *BindEndpoint) Bind(ctx context.Context, instanceID, bindingID string, d
 		InstanceID: instanceID,
 
 		CreatedAt: time.Now(),
+		ExpiresAt: time.Now().Add(time.Duration(int64(expirationSeconds)) * time.Second),
```
jaroslaw-pieszka, line 126: Are we sure the time of creation of this object is the same as the time of creation of the token?

internal/fixture/fixture.go
```diff
@@ -356,6 +356,7 @@ func FixBindingWithInstanceID(bindingID string, instanceID string) internal.Bind
 
 		CreatedAt: time.Now(),
 		UpdatedAt: time.Now().Add(time.Minute * 5),
+		ExpiresAt: time.Now().Add(time.Second * 600),
```
jaroslaw-pieszka, line 359: Please consider comment, there are people who do not get that 600 seconds is equal to 10 minutes.

internal/broker/bind_create.go
MarekMichali: I'm not sure, I will update this to check the time returned in the token.

jaroslaw-pieszka (approved):

PR #1322, opened 2024-10-15
Missing Bindings' Resources Removal

ralikio:
**Description**

- missed removal of service accounts, cluster roles and cluster role bindings during removal of a binding.

**Related issue(s)**

#284 #1198

### DIFF
--- a/cmd/broker/bind_create_test.go
+++ b/cmd/broker/bind_create_test.go
@@ -15,10 +15,10 @@ import (
 
 	"github.com/golang-jwt/jwt/v4"
 	"github.com/google/uuid"
+	brokerBindings "github.com/kyma-project/kyma-environment-broker/internal/broker/bindings"
 	"gopkg.in/yaml.v2"
 	rbacv1 "k8s.io/api/rbac/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	brokerBindings "github.com/kyma-project/kyma-environment-broker/internal/broker/bindings"
 
 	"code.cloudfoundry.org/lager"
 	"github.com/gorilla/mux"
@@ -451,8 +451,8 @@ func TestCreateBindingEndpoint(t *testing.T) {
 }
 
 func assertResourcesExistence(t *testing.T, k8sClient client.Client, bindingID string) {
-	name := "kyma-binding-" + bindingID	
-	namespace := "kyma-system"	
+	name := "kyma-binding-" + bindingID
+	namespace := "kyma-system"
 
 	serviceAccount := corev1.ServiceAccount{}
 	err := k8sClient.Get(context.Background(), client.ObjectKey{Name: name, Namespace: namespace}, &serviceAccount)
@@ -471,8 +471,8 @@ func assertResourcesExistence(t *testing.T, k8sClient client.Client, bindingID s
 }
 
 func assertServiceAccountsNotExists(t *testing.T, k8sClient client.Client, bindingID string) {
-	name := "kyma-binding-" + bindingID	
-	namespace := "kyma-namespace"	
+	name := "kyma-binding-" + bindingID
+	namespace := "kyma-namespace"
 
 	serviceAccount := corev1.ServiceAccount{}
 	err := k8sClient.Get(context.Background(), client.ObjectKey{Name: name, Namespace: namespace}, &serviceAccount)
--- a/cmd/broker/bind_create_test.go
+++ b/cmd/broker/bind_create_test.go
@@ -230,7 +230,7 @@ func TestCreateBindingEndpoint(t *testing.T) {
 
 		//// verify connectivity using kubeconfig from the generated binding
 		assertClusterAccess(t, "secret-to-check-first", binding)
-		assertRolesExistence(t, fmt.Sprintf(brokerBindings.BindingNameFormat, "binding-id"), binding)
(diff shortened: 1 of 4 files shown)


cmd/broker/bind_create_test.go
```diff
@@ -426,6 +450,43 @@ func TestCreateBindingEndpoint(t *testing.T) {
 	})
 }
 
+func assertResourcesExistence(t *testing.T, k8sClient client.Client, bindingID string) {
+	name := "kyma-binding-" + bindingID
+	namespace := "kyma-system"
```
jaroslaw-pieszka, line 455: Could you extract namespace to have it consistent across all test cases?

cmd/broker/bind_create_test.go
```diff
@@ -426,6 +450,43 @@ func TestCreateBindingEndpoint(t *testing.T) {
...
+	assert.NotNil(t, clusterRole)
+
+	clusterRoleBinding := rbacv1.ClusterRoleBinding{}
+	err = k8sClient.Get(context.Background(), client.ObjectKey{Name: name, Namespace: namespace}, &clusterRoleBinding)
+	assert.NoError(t, err)
+	assert.NotNil(t, clusterRoleBinding)
+}
+
+func assertServiceAccountsNotExists(t *testing.T, k8sClient client.Client, bindingID string) {
+	name := "kyma-binding-" + bindingID
```
jaroslaw-pieszka, line 474: In `clean code world` it would call for dedicated function forming the name in one single place.

internal/broker/bind_delete.go
```diff
@@ -25,10 +32,23 @@ func (b *UnbindEndpoint) Unbind(ctx context.Context, instanceID, bindingID strin
 	b.log.Infof("Unbind details: %+v", details)
 	b.log.Infof("Unbind asyncAllowed: %v", asyncAllowed)
 
-	err := b.bindingsStorage.Delete(instanceID, bindingID)
+	instance, err := b.instancesStorage.GetByID(instanceID)
+	switch {
+	case dberr.IsNotFound(err):
+		return domain.UnbindSpec{}, apiresponses.ErrInstanceDoesNotExist
```
jaroslaw-pieszka, line 38: What code caller gets in this case?

internal/broker/bind_delete.go
```diff
@@ -25,10 +32,23 @@ func (b *UnbindEndpoint) Unbind(ctx context.Context, instanceID, bindingID strin
...
+	case err != nil:
+		return domain.UnbindSpec{}, apiresponses.NewFailureResponse(fmt.Errorf("failed to get instance %s", instanceID), http.StatusInternalServerError, fmt.Sprintf("failed to get instance %s", instanceID))
```
jaroslaw-pieszka, line 40: As in previous question? What http code gets the caller?

internal/broker/bindings/bindings_manager.go
```diff
@@ -137,3 +139,36 @@ func (c *ServiceAccountBindingsManager) Create(ctx context.Context, instance *in
...
 }
+
+func (c *ServiceAccountBindingsManager) Delete(ctx context.Context, instance *internal.Instance, bindingID string) error {
+	clientset, err := c.clientProvider.K8sClientSetForRuntimeID(instance.RuntimeID)
+
+	if err != nil {
+		return fmt.Errorf("while creating a runtime client for binding creation: %v", err)
+	}
+
+	serviceBindingName := fmt.Sprintf("kyma-binding-%s", bindingID)
```
jaroslaw-pieszka, line 150: We could extract it as separate function - single point of truth - like with Runtime CR name, or Kyma CR name.
But this is nice to have, not must.

internal/broker/bind_delete.go
```diff
@@ -25,10 +32,23 @@ func (b *UnbindEndpoint) Unbind(ctx context.Context, instanceID, bindingID strin
...
+	err = b.bindingsManager.Delete(ctx, instance, bindingID)
+	if err != nil {
+		b.log.Errorf("Unbind error during removal of service account resources: %s", err)
+		return domain.UnbindSpec{}, err
+	}
 
+	err = b.bindingsStorage.Delete(instanceID, bindingID)
 	if err != nil {
-		b.log.Errorf("Unbind error: %s", err)
+		b.log.Errorf("Unbind error during removal of db entity: %s", err)
```
jaroslaw-pieszka, line 51:
```suggestion
		b.log.Errorf("Unbind error during removal of db entity: %v", err)
```

cmd/broker/bind_create_test.go
ralikio: Let me change it to function invocation instead of a method on one of bindings struct. It makes it easier to call such form from test because the tests operate on bindings structs that does not have id, but id is a part of binding name.

jaroslaw-pieszka: Cool.

internal/broker/bind_delete.go
ralikio: I believe it will be 404 https://github.com/openservicebrokerapi/servicebroker/blob/v2.17/spec.md#response-4

internal/broker/bind_delete.go
ralikio: 500

jaroslaw-pieszka (approved):

PR #1332, opened 2024-10-16
Fix binding e2e test

piotrmiskiewicz:
**Description**

- make possible (fix BrokerSuite test) to execute binding tests as e2e in cmd/broker

**Related issue(s)**

### DIFF
--- a/cmd/broker/suite_test.go
+++ b/cmd/broker/suite_test.go
@@ -926,8 +926,12 @@ func fixConfig() *Config {
 		Broker: broker.Config{
 			EnablePlans: []string{"azure", "trial", "aws", "own_cluster", "preview", "sap-converged-cloud", "gcp", "free"},
 			Binding: broker.BindingConfig{
-				Enabled:       true,
-				BindablePlans: []string{"aws", "azure"},
+				Enabled:              true,
+				BindablePlans:        []string{"aws", "azure"},
+				ExpirationSeconds:    100,
+				MaxExpirationSeconds: 1000,
+				MinExpirationSeconds: 1,
+				MaxBindingsCount:     2,
 			},
 			AllowUpdateExpiredInstanceWithContext: true,
 			KimConfig: broker.KimConfig{
--- a/cmd/broker/suite_test.go
+++ b/cmd/broker/suite_test.go
@@ -928,9 +928,9 @@ func fixConfig() *Config {
 			Binding: broker.BindingConfig{
 				Enabled:              true,
 				BindablePlans:        []string{"aws", "azure"},
-				ExpirationSeconds:    100,
+				ExpirationSeconds:    900,
 				MaxExpirationSeconds: 1000,
-				MinExpirationSeconds: 1,
+				MinExpirationSeconds: 600,
 				MaxBindingsCount:     2,
 			},
 			AllowUpdateExpiredInstanceWithContext: true,
--- a/internal/broker/bindings/bindings_manager.go
+++ b/internal/broker/bindings/bindings_manager.go
@@ -5,8 +5,7 @@ import (
 	"fmt"
 	"time"
 
-	v12 "k8s.io/client-go/kubernetes/typed/core/v1"
-	rbac "k8s.io/client-go/kubernetes/typed/rbac/v1"
+	"k8s.io/client-go/kubernetes"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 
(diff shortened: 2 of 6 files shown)


internal/broker/bindings/bindings_manager.go
```diff
@@ -31,7 +32,7 @@ type BindingsManager interface {
 }
 
 type ClientProvider interface {
-	K8sClientSetForRuntimeID(runtimeID string) (*kubernetes.Clientset, error)
+	K8sClientSetForRuntimeID(runtimeID string) (v12.CoreV1Interface, rbac.RbacV1Interface, error)
```
piotrmiskiewicz, line 35: the ClientSet is a struct, not interface, so it is not possible to provide fake ClientSet

cmd/broker/suite_test.go
```diff
@@ -926,8 +926,12 @@ func fixConfig() *Config {
 		Broker: broker.Config{
 			EnablePlans: []string{"azure", "trial", "aws", "own_cluster", "preview", "sap-converged-cloud", "gcp", "free"},
 			Binding: broker.BindingConfig{
-				Enabled:       true,
-				BindablePlans: []string{"aws", "azure"},
+				Enabled:              true,
+				BindablePlans:        []string{"aws", "azure"},
+				ExpirationSeconds:    100,
```
ralikio, line 931: Could we change the value to 600? That was the lowest possible value for Token Requests.

cmd/broker/suite_test.go
```diff
@@ -926,8 +926,12 @@ func fixConfig() *Config {
...
+				MaxExpirationSeconds: 1000,
+				MinExpirationSeconds: 1,
```
ralikio, line 933: Could we change the value to 600? That was the lowest possible value for Token Requests.

internal/broker/bindings/bindings_manager.go
```diff
@@ -89,12 +90,12 @@ func (c *ServiceAccountBindingsManager) Create(ctx context.Context, instance *in
 				},
 			},
 		}, mv1.CreateOptions{})
-
-	if err != nil && !apierrors.IsAlreadyExists(err) {
-		return "", time.Time{}, fmt.Errorf("while creating a cluster role: %v", err)
+	fmt.Printf("%v", e)
```
ralikio, line 93: Let's use logger implementation.

internal/broker/bindings/bindings_manager.go
```diff
@@ -73,13 +75,12 @@ func (c *ServiceAccountBindingsManager) Create(ctx context.Context, instance *in
 		return "", time.Time{}, fmt.Errorf("while creating a service account: %v", err)
 	}
 
-	_, err = clientset.RbacV1().ClusterRoles().Create(ctx,
+	_, e := rbacClient.ClusterRoles().Create(ctx,
```
ralikio, line 78: What is the reason that we change the name from "err" to "e"?

internal/broker/bindings/bindings_manager.go
piotrmiskiewicz: it was by accidend, I'll remove it

internal/broker/bindings/bindings_manager.go
piotrmiskiewicz: I was debugging and forgot to change it back, reverting...

internal/kubeconfig/provider.go
```diff
@@ -106,9 +114,42 @@ func (p *FakeProvider) K8sClientForRuntimeID(_ string) (client.Client, error) {
...
+`), nil
+}
+
+func (p *FakeProvider) K8sClientSetForRuntimeID(runtimeID string) (v12.CoreV1Interface, rbac.RbacV1Interface, error) {
+	return p.coreClient, p.rbacClient, nil
 }
 
-func (p *FakeProvider) K8sClientSetForRuntimeID(runtimeID string) (*kubernetes.Clientset, error) {
-	return nil, fmt.Errorf("not implemented")
+func createFakeClients() (v12.CoreV1Interface, rbac.RbacV1Interface) {
```
ralikio, line 143: What do you think about using fake clientset from client-go?

internal/kubeconfig/provider.go
```diff
@@ -4,11 +4,16 @@ import (
 	"context"
 	"fmt"
 
+	machineryv1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v12 "k8s.io/client-go/kubernetes/typed/core/v1"
```
jaroslaw-pieszka, line 8: Could we use other alias? `typedv1`?

internal/broker/bindings/bindings_manager.go
ralikio: Can we use the following interface instead of two separate interfaces: https://github.com/kubernetes/client-go/blob/235936510e4ea63c889836f7d6b06ddcf6dfa08e/kubernetes/clientset.go#L83

internal/kubeconfig/provider.go
piotrmiskiewicz: changed to kubernetes.Interface

ralikio (approved):

PR #1345, opened 2024-10-17
Adjust HTTP codes for bindings PUT

MarekMichali:
**Description**

- Return 409 when trying to create a binding with existing ID but different parameters
- Return 200 when trying to create a binding with existing ID and the same parameters
- Return 400 when trying to insert to DB a binding with ID that already exists

**Related issue(s)**

See #1352

### DIFF
--- a/cmd/broker/bind_create_test.go
+++ b/cmd/broker/bind_create_test.go
@@ -476,42 +476,6 @@ func TestCreateBindingEndpoint(t *testing.T) {
 		//then
 		require.Equal(t, http.StatusBadRequest, response.StatusCode)
 	})
-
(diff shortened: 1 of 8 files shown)


cmd/broker/bind_create_test.go
```diff
@@ -476,6 +476,42 @@ func TestCreateBindingEndpoint(t *testing.T) {
 		//then
 		require.Equal(t, http.StatusBadRequest, response.StatusCode)
 	})
+
```
piotrmiskiewicz, line 479: those test cases does not need 'envtest' (envtest increases the time of the execution). Please move those test cases to binding_test or even to internal/broker  - you need to decide

piotrmiskiewicz (approved):

PR #1355, opened 2024-10-17
Return 404 from GET Bindings for Expired Bindings

ralikio:
**Description**

- returning 404 error code from API GET bindings in case of binding expiration,
- test case for above.

**Related issue(s)**

#284


internal/broker/bind_get_test.go
```diff
@@ -0,0 +1,44 @@
...
+		}
+
+		// when
+		_, err := endpoint.GetBinding(context.Background(), "test-instance-id", "test-binding-id", domain.FetchBindingDetails{})
+
+		// then
+		require.NotNil(t, err)
+		apiErr, ok := err.(*apiresponses.FailureResponse)
+		require.True(t, ok)
+		require.Equal(t, http.StatusNotFound, apiErr.ValidatedStatusCode(nil))
```
jaroslaw-pieszka, line 41: Could you assert for the error message as well (`Binding expired`)

internal/broker/bind_get_test.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/pivotal-cf/brokerapi/v8/domain/apiresponses"
+	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/require"
+)
+
+func TestGetBinding(t *testing.T) {
+
+	t.Run("should return 404 code for the expired binding", func(t *testing.T) {
+		// given
+		mockBindings := new(mocks.Bindings)
```
piotrmiskiewicz, line 21: just a question, why just not use in-memory implementation?

ralikio: With mocks I am guaranteed the bindings "Get" method is called.

piotrmiskiewicz: We are using in memory implementation in other unit test which works fine, I don't like to introduce a different solution which does not give us any special value. Such pattern is also used in fake kubernetes client. Mocking database can be risky. 
The title of the test is "should return 404 code for the expired binding", not "should call 'get' method", it is not important in my opinion.

ralikio: I still find value in asserting execution. In order to meet the argument of unnecessary test modifications when interface changes and mock regeneration is required I would suggest to treat interface as something that rarely should change or rely on new interfaces in new logic.

Done as requested though. Let's discuss offline what are the possibilities to use spies or sth else to assert method executions in our project.

jaroslaw-pieszka: It's not about interface changing but implementation changing. We relied on implementation which in this case I found excessive, or even inappropriate.

ralikio: No, that is not true. The test contained assertion on logic so that in the future if our "black box" called bindings multiple times we would know that it is not aligned with initial assumptions. Think about it not as an assertion if specific instruction that the unit uses but as requirement introduced because unit relies on external dependency in the form of "Bindings" interface. In the current form, we could remove Bindings dependency, return 404 with correct message from the unit and the test would be green.

jaroslaw-pieszka: I beg to disagree. Test assumed that implementation calls `Get` method once. This is not part of contract.

jaroslaw-pieszka: Assuming that we test contract and use black-box testing, we do not need to know that there is any external dependency, we do not need to know how it is implemented, how many invocations of any kind is made.
We know only, that if the binding in question is expired, we get 404.

piotrmiskiewicz (approved):

internal/broker/bind_get_test.go
piotrmiskiewicz: We can testing if there is only one Get call on repository, but... then we should test if the binding manager deletes... deployments or something else by accident. We cannot test everything! such tests would be a problem, not a help

Issue #1349, opened 2024-10-17
SB: Correct Bad Request Responses


ralikio:
Currently create endpoint returns 500 on unmarshal error if any parameter in input json is incorrect.

**AC** 
- [x] Correct this behaviour by returning BAD Request in such cases 
- [x] add details description about what fields are incorrect
- [x] Description should be returned in the response with details information about which field has failed

PR #1360, opened 2024-10-18
Handle incorrect binding parameters

KsaweryZietara:
**Description**

- return bad request status for incorrect parameters,
- remove unused dummy credentials.

**Related issue(s)**
See also #1349

### DIFF
--- a/cmd/broker/binding_test.go
+++ b/cmd/broker/binding_test.go
@@ -52,7 +52,7 @@ func TestBinding(t *testing.T) {
 	suite.Log(string(b))
 	suite.Log(resp.Status)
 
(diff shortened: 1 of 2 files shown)


cmd/broker/binding_test.go
```diff
@@ -49,4 +51,16 @@ func TestBinding(t *testing.T) {
 	b, _ = io.ReadAll(respRuntimes.Body)
 	suite.Log(string(b))
 	suite.Log(resp.Status)
+
+	t.Run("should return 400 for incorrect parameters", func(t *testing.T) {
```
jaroslaw-pieszka, line 55: Please add comment or change name to clarify, that incorrect means string instead of int.

jaroslaw-pieszka (approved):

Issue #1364, opened 2024-10-18
Service Binding Insert Order


ralikio:
**AC**
- [x] 1) Make insert at begginng
- [ ] 2) Make update at the end if kubeconfig ready

Issue from comment: https://github.com/kyma-project/kyma-environment-broker/issues/284#issuecomment-2416064563

piotrmiskiewicz:
```
1. Insert binding into the storage (kubeconfig="", expirationAt=now+expirationSeconds)

2. Create resoruces and token request

3. Update binding with kubeconfig, expirationAt=tokenRequest.ExpirationAt
```

PR #1362, opened 2024-10-18
SB: Make sure Service Bindings do not do not block Service Instance deletion

jaroslaw-pieszka:
**Description**

Changes proposed in this pull request:

- Testing if deprovisioning of Service Instance with Service Bindings is successful
- Extending `Unbind` endpoint to handle suspended, deprovisioned, being deprovisioned Service Instances as well
- Extending `GetBinding` endpoint to handle suspended, deprovisioned instances, being deprovisioned Service Instances
**Related issue(s)**

#1199

PR #1366, opened 2024-10-21
Corrected Binding Insert Order

ralikio:
**Description**

- put insert before K8S resources creation in binding creation process,
- test that rely on nil providers to make sure that before panic occurs the test created a binding,
- updating expires at and a kubeconfig after its generation,
- asserting that expires at is set even when failed occured.

**Related issue(s)**

#1364


internal/broker/bind_create.go
```diff
@@ -163,23 +163,23 @@ func (b *BindEndpoint) Bind(ctx context.Context, instanceID, bindingID string, d
...
-	}
-
-	binding.ExpiresAt = expiresAt
-	binding.Kubeconfig = kubeconfig
 
 	err = b.bindingsStorage.Insert(binding)
 	if err != nil {
 		message := fmt.Sprintf("failed to insert Kyma binding into storage: %s", err)
 		return domain.Binding{}, apiresponses.NewFailureResponse(fmt.Errorf(message), http.StatusInternalServerError, message)
+	}
```
jaroslaw-pieszka, line 171: We do not set expiresAt before DB Insert.

internal/broker/bind_create_test.go
```diff
@@ -3,9 +3,120 @@ package broker
...
+	User struct {
+		Token string `yaml:"token"`
+	} `yaml:"user"`
+}
+
+const (
+	instanceID1          = "1"
+	instanceID2          = "2"
+	instanceID3          = "max-bindings"
+	expirationSeconds    = 10000
```
jaroslaw-pieszka, line 37: Why expirationSeconds is greater than maxExpirationSeconds? I bet there is a reason.

ralikio: Params were unnecessary and copied from other test. Corrected by removing them in https://github.com/kyma-project/kyma-environment-broker/pull/1366/commits/805c1ac8e7de479a4d8a554ce2ab2a8b723b44e0.

internal/broker/bind_create_test.go
```diff
@@ -11,8 +11,104 @@ import (
...
+			}
+
+			binding, err := db.Bindings().Get(instanceID1, "binding-id")
+			require.NoError(t, err)
+			require.Equal(t, instanceID1, binding.InstanceID)
+			require.Equal(t, "binding-id", binding.ID)
+		}()
+
+		// given
+		_, err := db.Bindings().Get(instanceID1, "binding-id")
```
piotrmiskiewicz, line 100: why there is a binding before "when"? I thought we are testing scenario that we want to put a binding and there is an error in the middle of Binding creation process. Currently, the scenario is about max number of bindings reached... then require.NoError in line 94 fails

piotrmiskiewicz: I don't want to force my idea, just to see another idea:
```
type provider struct {
}

func (p *provider) K8sClientSetForRuntimeID(runtimeID string) (kubernetes.Interface, error) {
	c := fake.NewSimpleClientset()
	c.PrependReactor("create", "serviceaccounts", func(action k8stesting.Action) (bool, runtime.Object, error) {
		return true, nil, fmt.Errorf("error")
	})
	return c, nil
}
```

then the test without recover:

```
_, err = db.Bindings().Get(instanceID1, "binding-id")
	require.Error(t, err)
	require.True(t, dberr.IsNotFound(err))

	// when
	_, _ = bindEndpoint.Bind(context.Background(), instanceID1, "binding-id", domain.BindDetails{
		ServiceID: "123",
		PlanID:    fixture.PlanId,
	}, false)

	// then
	binding, err := db.Bindings().Get(instanceID1, "binding-id")
	require.NoError(t, err)
	require.Equal(t, instanceID1, binding.InstanceID)
	require.Equal(t, "binding-id", binding.ID)

```

internal/broker/bind_create_test.go
```diff
@@ -11,8 +11,104 @@ import (
...
+		TimestampFormat: time.RFC3339Nano,
+	})
+
+	brokerLogger := lager.NewLogger("test")
+	brokerLogger.RegisterSink(lager.NewWriterSink(logs.Writer(), lager.DEBUG))
+
+	//// schema
+
+	//// database
+	db := storage.NewMemoryStorage()
```
jaroslaw-pieszka, line 61: If this PR is to be merged after #1362 - we will need some adjustments here. We will need operation inserted in `operations` table for each instance, because GET endpoint handler checks last operation.

internal/storage/driver/memory/binding.go
```diff
@@ -40,6 +40,18 @@ func (s *Binding) Insert(binding *internal.Binding) error {
 	return nil
 }
 
+func (s *Binding) Update(binding *internal.Binding) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	if _, found := s.data[binding.ID]; !found {
```
jaroslaw-pieszka, line 47: Should we put a comment here, that this behaviour is different than this of postgres?

internal/broker/bind_create.go
jaroslaw-pieszka: Addressed.

internal/storage/driver/memory/binding.go
ralikio: Corrected in https://github.com/kyma-project/kyma-environment-broker/pull/1366/commits/8700a18713ab50158aab98877d675ba8b8da66aa

internal/storage/postsql/write.go
```diff
@@ -56,6 +56,21 @@ func (ws writeSession) InsertBinding(binding dbmodel.BindingDTO) dberr.Error {
 	return nil
 }
 
+func (ws writeSession) UpdateBinding(binding dbmodel.BindingDTO) dberr.Error {
```
jaroslaw-pieszka, line 59: I wonder if we should add `updated_at` field to `bindings` table if we allow update operations.

ralikio: For now we have expiration_seconds plus expired_at that could be used to calculate that value. Lets' discuss it outside of this PR.

internal/storage/driver/memory/binding.go
```diff
@@ -32,14 +32,26 @@ func (s *Binding) Insert(binding *internal.Binding) error {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
-	if _, found := s.data[binding.ID]; found {
+	if foundBinding, found := s.data[binding.ID]; found && binding.InstanceID == foundBinding.InstanceID {
 		return dberr.AlreadyExists("binding with id %s already exist", binding.ID)
```
jaroslaw-pieszka, line 36:
```suggestion
		return dberr.AlreadyExists("binding with id %s already exists", binding.ID)
```

internal/storage/driver/memory/binding.go
```diff
@@ -32,14 +32,26 @@ func (s *Binding) Insert(binding *internal.Binding) error {
...
 	s.data[binding.ID] = *binding
 
 	return nil
 }
 
+func (s *Binding) Update(binding *internal.Binding) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	if foundBinding, found := s.data[binding.ID]; !found || binding.InstanceID != foundBinding.InstanceID {
```
jaroslaw-pieszka, line 47:
```suggestion
	if foundBinding, found := s.data[binding.ID]; !(found && binding.InstanceID == foundBinding.InstanceID) {
```
I am not sure which is better.

jaroslaw-pieszka: First is definitely shorter.

jaroslaw-pieszka (approved):

Issue #1196, opened 2024-09-26
SB: Make our swagger file up to date with Service Bindings OSB APIs


PK85:
Subtask of the EPIC: https://github.com/kyma-project/kyma-environment-broker/issues/284

**AC**
- [x] make swagger for SBs up to date, see: https://kyma-env-broker.cp.dev.kyma.cloud.sap/

PR #1376, opened 2024-10-23
Update swagger with bindings

MarekMichali:
**Description**

- Update PUT, GET, DELETE for bindings

**Related issue(s)**

See #1196

### DIFF
--- a/resources/keb/files/swagger.yaml
+++ b/resources/keb/files/swagger.yaml
@@ -681,11 +681,11 @@ paths:
 
   /oauth/v2/service_instances/{instance_id}/service_bindings/{binding_id}:
     put:
-      summary: generation of a service binding
+      summary: create a service binding
       security:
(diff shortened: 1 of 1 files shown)


resources/keb/files/swagger.yaml
```diff
@@ -1354,12 +1225,18 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Error'
+        '500':
+          description: Internal server error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
```
MarekMichali, line 1233: I'm not 100% sure if we should use the generic `Error` or create an appropriate error message for each code. For now I decided to use the generic `Error` as we have the same approach in the `Instances` section.

jaroslaw-pieszka (approved):

PR #1382, opened 2024-10-23
SB: documentation updated

jaroslaw-pieszka:
**Description**

- Status codes for DELETE in case of suspension/deprovisioning, non-existing instance
- Status codes for GET in case of suspension, non-existing instance

**Related issue(s)**

### DIFF
--- a/docs/user/05-60-kyma-bindings.md
+++ b/docs/user/05-60-kyma-bindings.md
@@ -37,7 +37,7 @@ The Broker returns the `200 OK` status code with the kubeconfig in the response 
 
 ## Unbinding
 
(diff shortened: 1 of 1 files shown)


MarekMichali:
The whole PUT request should also be updated. It still has this part `accepts_incomplete=false&service_id={{service_id}}&plan_id={{plan_id}}` which should be removed.

MarekMichali:
Maybe we should also mention in the PUT, that this request also returns metadata with the expires_at field.

MarekMichali (approved):

IwonaLanger (approved):

Issue #1378, opened 2024-10-23
SB: Additional logging


ralikio:
Additional logging for service bindings - verification + addition:
- [x] successful creation of binding for specific instance,
- [x] get called for expired binding,
- [x] binding created but limit reach,
- [x] happy binding removal,
- [x] ...

MarekMichali:
Added also logs when the token creation fails.

PR #1387, opened 2024-10-25
Extend bindings logs

MarekMichali:
**Description**

- More logs

**Related issue(s)**

See #1378


internal/broker/bind_delete.go
```diff
@@ -73,6 +73,7 @@ func (b *UnbindEndpoint) Unbind(ctx context.Context, instanceID, bindingID strin
 		b.log.Errorf("Unbind error during removal of db entity: %v", err)
 		return domain.UnbindSpec{}, apiresponses.NewFailureResponse(fmt.Errorf("failed to delete binding resources for binding %s and instance %s: %v", bindingID, instanceID, err), http.StatusInternalServerError, fmt.Sprintf("failed to delete resources for binding %s and instance %s: %v", bindingID, instanceID, err))
 	}
+	b.log.Infof("Successfully removed binding %s for instance %s", bindingID, instanceID)
```
ralikio, line 76: Could we change "Successfully" to lower case?

ralikio (approved):

internal/broker/bind_delete.go
MarekMichali: We decided to leave it as it is for now.

Issue #1377, opened 2024-10-23
SB: Bindings Creation Timeout


ralikio:
Please introduce internal overall timeout set to 15 seconds for whole binding creation.

**AC**
- [x] configurable timeout set with the usage of KEB config, default value set to 15 seconds

PR #1388, opened 2024-10-25
Timeout of creation Service Binding

ukff:
**Changes proposed in this pull request**

- Add expiration logic to Bind create endpoint. Process will fail if it will now finish in configurable time.

**Related issue(s)**

https://github.com/kyma-project/kyma-environment-broker/issues/1377


internal/broker/server.go
```diff
@@ -11,8 +13,16 @@ import (
 	"github.com/pivotal-cf/brokerapi/v8/middlewares"
 )
 
+type CreateBindingHandler struct {
+	handler func(w http.ResponseWriter, req *http.Request)
+}
+
+func (h CreateBindingHandler) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
```
jaroslaw-pieszka, line 20: Where is this `func` used?

jaroslaw-pieszka (approved):

PR #1396, opened 2024-10-30
Removal of Unnecessary Operation

ralikio:
**Description**

Delete operation is not necessary at this point of binding removal since binding removal should pickup such leftovers.

**Related issue(s)**

#284


ralikio:
Reopening PR to check if all tests pass.

PR #1423, opened 2024-11-07
Create and promote release pipeline

KsaweryZietara:
**Description**

- add new pipeline which will create release and promote release tag from this action.


.github/workflows/create-release.yaml
```diff
@@ -23,6 +23,21 @@ on:
...
         description: 'Do not publish'
         default: false
+  workflow_call:
+    inputs:
+      name:
+        type: string
+        description: 'Create release'
+        default: ""
+        required: true
+      sec-file-and-chart:
```
MarekMichali, line 33: I would rename it to `skip-sec-file-and-chart` or something like that, because now if sec-file-and-chart is false it may suggest that we don't want to create the bump.

MarekMichali (approved):

PR #1447, opened 2024-11-13
Add more check for PR's merged during release

MarekMichali:
**Description**

This checks if a PR has been merged after the first PR merge check and before the bumps are merged. 
This checks also if a PR has been merged after the second PR merge check and before the draft release and tag are created. 
This covers the corner case scenario where there is about a ~1 s window in the whole release where we could merge a PR and not detect it when needed.

**Related issue(s)**

### DIFF
--- a/.github/workflows/create-release.yaml
+++ b/.github/workflows/create-release.yaml
@@ -218,7 +218,7 @@ jobs:
 
       - name: Check if any PR was merged during release
         run: |
-          COMMIT_SHA=$(gh api repos/${{ env.KYMA_ENVIRONMENT_BROKER_REPO }}/commits/main -q '.sha')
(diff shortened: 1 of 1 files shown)


.github/workflows/create-release.yaml
```diff
@@ -268,6 +268,16 @@ jobs:
             echo "Step skipped"
           fi
 
+      - name: Check if any PR have been merged before merging the bump PR
+        run: |
+          COMMIT_SHA=$(gh api repos/${{ env.KYMA_ENVIRONMENT_BROKER_REPO }}/commits -q '.[1].sha')
```
MarekMichali, line 273: this check the commit before the PR bump

ralikio (approved):

PR #1473, opened 2024-11-21
Additional Logging in Workers

ralikio:
**Description**

- additional logging in the queue to give more insides on its behaviour,
- health check that warns if a worker has not been executing for configurable amount of time.

**Related issue(s)**


internal/process/queue_test.go
```diff
@@ -0,0 +1,100 @@
...
+		queue := NewQueue(&StdExecutor{logger: func(msg string) { t.Log(msg) }}, logger, "test", 10*time.Second, 10*time.Second)
+		queue.Run(cancelContext.Done(), 1)
+
+		queue.AddAfter("processId2", 0)
+		queue.Add("processId")
+		queue.SpeedUp(1)
+
+		for queue.queue.Len() > 0 {
+			// wait for the queue to be drained:
+			time.Sleep(500 * time.Millisecond)
```
piotrmiskiewicz, line 43: 500 millisecond is too long, try to reduce it

piotrmiskiewicz: you can even directly wait until the executer is executed, something like:
```
var waitGroup sync.WaitGroup
		cancelContext, cancel := context.WithCancel(context.Background())
		queue := NewQueue(&StdExecutor{logger: func(msg string) {
			t.Log(msg)
			waitGroup.Done()
		}}, logger, "test", 10*time.Second, 10*time.Second)
		queue.Run(cancelContext.Done(), 1)

		waitGroup.Add(2)
		queue.AddAfter("processId2", 0)
		queue.Add("processId")
		queue.SpeedUp(1)

		waitGroup.Wait()

		queue.ShutDown()
		cancel()
		queue.waitGroup.Wait()
```

cmd/broker/deprovisioning.go
```diff
@@ -97,7 +97,7 @@ func NewDeprovisioningProcessingQueue(ctx context.Context, workersAmount int, de
 		}
 	}
 
-	queue := process.NewQueue(deprovisionManager, logs)
+	queue := process.NewQueue(deprovisionManager, logs, "deprovisioning-queue", cfg.Broker.WorkerHealthCheckWarnInterval, cfg.Broker.WorkerHealthCheckInterval)
```
piotrmiskiewicz, line 100: nit: "privisioning-queue" -> "provisioning" - we don't need to name the queue with word "queue", we know it :)

piotrmiskiewicz: the same for provisioning and an update

internal/process/queue.go
```diff
@@ -15,78 +17,111 @@ type Executor interface {
...
+	q.log.Infof("item %s will be added to the queue %s after duration of %d, queue length is %d", processId, q.name, duration, q.queue.Len())
 }
 
 func (q *Queue) ShutDown() {
+	q.log.Infof("shutting down the queue, queue length is %d", q.queue.Len())
 	q.queue.ShutDown()
 }
 
 func (q *Queue) Run(stop <-chan struct{}, workersAmount int) {
+	q.log.Infof("starting workers, queue length is %d", q.queue.Len())
```
jaroslaw-pieszka, line 63: Could we log workersAmount? When starting the queue length would be 0 unless we allow to insert elements before start.

internal/process/queue.go
```diff
@@ -96,8 +131,43 @@ func (q *Queue) worker(queue workqueue.RateLimitingInterface, process func(key s
...
+	for name, lastTime := range q.workerExecutionTimes {
+		timeSinceLastExecution := time.Since(lastTime)
+
+		buffer.WriteString(fmt.Sprintf(", [worker %s, last execution time: %s, since last execution: %s]", name, lastTime, timeSinceLastExecution))
+	}
+
+	healthCheckLog.Info(buffer.String())
+
+	for name, lastTime := range q.workerExecutionTimes {
+		timeSinceLastExecution := time.Since(lastTime)
```
jaroslaw-pieszka, line 168: The same loop again. Code duplication.

internal/process/queue.go
```diff
@@ -96,8 +131,43 @@ func (q *Queue) worker(queue workqueue.RateLimitingInterface, process func(key s
...
 		}
 	}
 }
+
+func (q *Queue) logAndUpdateWorkerTimes(key string, name string, log logrus.FieldLogger) {
+	// log time
+	now := time.Now()
+	lastTime, ok := q.workerExecutionTimes[name]
+	if ok {
+		log.Infof("execution - worker %s last execution time %s, executed after %s", name, lastTime, now.Sub(lastTime))
```
jaroslaw-pieszka, line 147: Is this specifier appropriate for time value?

internal/process/queue.go
ralikio: It is by design. I do not want warn messages appearing before worker times are logged.

internal/process/queue.go
```diff
@@ -15,78 +17,111 @@ type Executor interface {
...
 				key, shutdown := queue.Get()
 				if shutdown {
+					log.Infof("shutting down")
 					return true
 				}
+
 				id := key.(string)
 				log = log.WithField("operationID", id)
+				log.Infof("about to process item %s, queue length is %d", id, q.queue.Len())
+				q.logAndUpdateWorkerTimes(key.(string), nameId, log)
```
jaroslaw-pieszka, line 112: Since `Get` blocks when queue is empty we will get false warnings/alerts .

internal/process/queue.go
```diff
@@ -96,8 +131,43 @@ func (q *Queue) worker(queue workqueue.RateLimitingInterface, process func(key s
...
+	lastTime, ok := q.workerExecutionTimes[name]
+	if ok {
+		log.Infof("execution - worker %s last execution time %s, executed after %s", name, lastTime, now.Sub(lastTime))
+	}
+	q.workerExecutionTimes[name] = now
+
+	log.Infof("processing item %s, queue length is %d", key, q.queue.Len())
+}
+
+func (q *Queue) HealthCheck() {
```
jaroslaw-pieszka, line 154: This `health check` checks interval between `now` and the last time workers processed the queue element. 

What is the intended purpose? What should this health check detect?
Would it detect long processing?
Would it detect worker not being called by called?

internal/process/queue.go
ralikio: Right, I will change "Warn" to "Info"

jaroslaw-pieszka: I do not get the point of this change.

internal/process/queue.go
jaroslaw-pieszka: One loop and two buffers would suffice.

internal/process/queue.go
ralikio: Yes, the logs are formatted as follows: "executed after 161.541µs"

internal/process/queue.go
ralikio: This is to detect how many workers are stuck, whatever a reason may be.

Would it detect long processing? - yes
Would it detect worker not being called by wait.Until? - yes
What would happen if queue is empty for some time? - that is a good point, which made me change warn level to info

Please, treat as information only

internal/process/queue.go
jaroslaw-pieszka: Given the number of workers it is not the issue in terms of performance.
Clean code zealots would object.

internal/process/queue.go
jaroslaw-pieszka: What I mean is that this is not a health check. Or oversensitive health check.

ralikio: We can change the name if you want. Do you have a proposal?

jaroslaw-pieszka: `logWorkerExecutionTimes`

internal/process/queue.go
jaroslaw-pieszka: My point is that it detects long time between subsequent `logAndUpdateWorkerTimes` calls.
The reason could be:
- empty queue
- worker not being selected on blocking `Get` (other worker selected)
- long processing 
- not being called by `wait.Until`

internal/process/queue_test.go
```diff
@@ -0,0 +1,106 @@
...
+
+		waitGroup.Add(2)
+		queue.AddAfter("processId2", 0)
+		queue.Add("processId")
+		queue.SpeedUp(1)
+
+		waitGroup.Wait()
+
+		queue.ShutDown()
+		cancel()
```
jaroslaw-pieszka, line 51: Is this necessary?

internal/process/queue_test.go
```diff
@@ -0,0 +1,106 @@
...
+		queue.waitGroup.Wait()
```
jaroslaw-pieszka, line 52: Is this necessary?

internal/process/queue_test.go
ralikio: Yes, I want to check logs after complete queue shutdown and because current implementation that is the way to do it.

internal/process/queue_test.go
ralikio: Duplicate

internal/process/queue_test.go
```diff
@@ -0,0 +1,106 @@
...
+		queue := NewQueue(&StdExecutor{logger: func(msg string) { 
+			t.Log(msg) 
+			waitGroup.Done()
+		}}, logger, "test", 10*time.Millisecond, 10*time.Millisecond)
+		queue.Run(cancelContext.Done(), 1)
+
+		waitGroup.Add(2)
+		queue.AddAfter("processId2", 0)
+		queue.Add("processId")
+		queue.SpeedUp(1)
```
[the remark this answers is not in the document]
ralikio: To go into speed up method and log the message inside of it.

internal/process/queue.go
ralikio: I really need this line to quickly find what was going on. In all cases mentioned by you if Info is logged it means that we either could adjust workers number or we have some unexpected behaviour.

internal/process/queue_test.go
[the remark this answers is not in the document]
jaroslaw-pieszka: The point is this is default value. 
So are you trying to prepare for the change of default value?

internal/process/queue_test.go
jaroslaw-pieszka: So logs will not be available before shutdown?

ralikio: I think so. Because we have a `for !exit` loop that requires a shutdown to be called. Am I missing something?

piotrmiskiewicz (approved):

internal/process/queue_test.go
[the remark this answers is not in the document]
jaroslaw-pieszka: I got it. You want to log it.

internal/process/queue_test.go
jaroslaw-pieszka: Where do we listen for cancelation?

jaroslaw-pieszka (approved):

Issue #1464, opened 2024-11-19
Env Cleanup job works on shootNames not runtimeIds


PK85:
https://sap-btp.slack.com/archives/C01LGCBS196/p1728903306048659?thread_ts=1728633328.899129&cid=C01LGCBS196

`The way the cleanup job works is that it selects all the shoots from Gardener that don't have this do-not-delete label set to true. Then it logs the shoot id with an info if the shoot is older than 24 hours. The shoots older than 24 hours are deleted. The job takes the runtimeID from the annotations in the shoot CR on Gardener and uses it to send the deprovision request.`

Summary: We suspect that a shoot CR older than 24 hours had in annotations a runtimeID of an another shoot that shouldn't be deprovisioned. We are missing logs to 100% confirm that.

**AC**
- [x] Before sending deprovision log a tuple {shootID, runtimeID, instanceID}
- [x] to not have issues that someone messed up labels in the shoot CR, we should calculate list for deletion by using ShootNames
  - [x] do not use any annotations or labels, we need to use shootName for KEB and RuntimeCR

PR #1502, opened 2024-11-28
Env cleanup job works on shoot names

KsaweryZietara:
**Description**

We shouldn't calculate lists for deletion based on labels or annotation, because user can change them. Instead, check if instance for shoot name exist in the database, if not, check if runtime custom resource was created for this shoot name, otherwise delete shoot in Gardener.  We have to fetch all runtime custom resources, because field selector for shoot name is not supported.

- calculate lists for deletion based on shoot names,
- extend logging,
- adjust tests.

**Related issue(s)**
See also #1464

### DIFF
--- a/internal/environmentscleanup/service.go
+++ b/internal/environmentscleanup/service.go
@@ -287,10 +287,6 @@ func (s *Service) triggerEnvironmentDeprovisioning(instance internal.Instance) e
 func (s *Service) cleanUpRuntimeCRs(runtimeCRs []runtime) error {
 	s.logger.Infof("RuntimeCRs to process: %+v", runtimeCRs)
 
-	if len(runtimeCRs) == 0 {
-		return nil
-	}
-
 	var result *multierror.Error
 
 	for _, runtime := range runtimeCRs {
--- a/internal/environmentscleanup/service.go
+++ b/internal/environmentscleanup/service.go
@@ -47,8 +47,6 @@ type Service struct {
 	logger          *log.Logger
 	MaxShootAge     time.Duration
 	LabelSelector   string
-	instances       map[string]internal.Instance
-	runtimeCRs      map[string]imv1.Runtime
 }
 
 type runtime struct {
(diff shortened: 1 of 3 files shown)


internal/environmentscleanup/service.go
```diff
@@ -189,39 +192,30 @@ func (s *Service) getStaleRuntimesByShoots(labelSelector string) ([]runtime, []u
 		}
 
 		log.Infof("Shoot %q is older than %f hours with age: %f hours", shoot.GetName(), s.MaxShootAge.Hours(), shootAge.Hours())
-		staleRuntime, err := s.shootToRuntime(shoot)
-		if err != nil {
-			s.logger.Infof("found a shoot without kcp labels: %v", shoot.GetName())
-			shoots = append(shoots, shoot)
+
+		instances, _, _, _ := s.instanceStorage.List(dbmodel.InstanceFilter{Shoots: []string{shoot.GetName()}})
```
jaroslaw-pieszka, line 196: There is no index based on shoot in `instances` table so we will have len(shootList.Items) full scans in the table.
Would it be the issue?

internal/environmentscleanup/service.go
```diff
@@ -292,30 +284,53 @@ func (s *Service) triggerEnvironmentDeprovisioning(instance internal.Instance) e
...
-		for _, instance := range kebInstancesToDelete {
-			if instance.RuntimeID == runtimeID {
-				return true
-			}
-		}
+func (s *Service) cleanUpRuntimeCRs(runtimeCRs []runtime) error {
+	s.logger.Infof("RuntimeCRs to process: %+v", runtimeCRs)
 
-		return false
+	if len(runtimeCRs) == 0 {
```
jaroslaw-pieszka, line 290: Is it really necessary? Would not `for` loop suffice?

internal/environmentscleanup/service.go
```diff
@@ -292,30 +284,53 @@ func (s *Service) triggerEnvironmentDeprovisioning(instance internal.Instance) e
...
 			}
+			return strings.Join(s, ", ")
 		}
 	}
 
-	return result
+	return result.ErrorOrNil()
+}
+
+func (s *Service) getRuntimeCR(shootName string) (runtime, error) {
```
jaroslaw-pieszka, line 317: We fetch all Runtime CRs just to select one.
See comment in the place of invocation.

internal/environmentscleanup/service.go
KsaweryZietara: There are around 100 instances on dev, so I think it won't be an issue. We use this function to fetch runtimes in `kcp rt` and it's quite fast.

internal/environmentscleanup/service.go
```diff
@@ -189,39 +192,30 @@ func (s *Service) getStaleRuntimesByShoots(labelSelector string) ([]runtime, []u
...
-}
-
-func (s *Service) shootToRuntime(st unstructured.Unstructured) (*runtime, error) {
-	shoot := gardener.Shoot{Unstructured: st}
-	runtimeID, ok := shoot.GetAnnotations()[shootAnnotationRuntimeId]
-	if !ok {
-		runtimeID, ok = shoot.GetAnnotations()[shootAnnotationInfrastructureManagerRuntimeId]
-		if !ok {
-			return nil, fmt.Errorf("shoot %q has no runtime-id annotation", shoot.GetName())
+		runtimeCR, err := s.getRuntimeCR(shoot.GetName())
```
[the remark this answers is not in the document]
KsaweryZietara: Typically there are less than 10 instances to remove, and most of them are provisioned using KEB (in this case this function is not called). I can't fetch by shoot name, because field selector is not supported.

internal/environmentscleanup/service.go
KsaweryZietara: I will remove it.

internal/environmentscleanup/service.go
jaroslaw-pieszka: This job runs only on DEV. I forgot about it.

internal/environmentscleanup/service.go
```diff
@@ -168,16 +170,36 @@ func (s *Service) cleanupShoots(shoots []unstructured.Unstructured) error {
...
 	opts := v1.ListOptions{
 		LabelSelector: labelSelector,
 	}
 	shootList, err := s.gardenerService.List(context.Background(), opts)
 	if err != nil {
-		return []runtime{}, []unstructured.Unstructured{}, fmt.Errorf("while listing Gardener shoots: %w", err)
+		return []runtime{}, []runtime{}, []unstructured.Unstructured{}, fmt.Errorf("while listing Gardener shoots: %w", err)
+	}
+
+	instancesMap := make(map[string]internal.Instance)
```
jaroslaw-pieszka, line 182: To decrease memory footprint consider map[shoot]->runtime instead of instance.

jaroslaw-pieszka (approved):

Issue #1469, opened 2024-11-20
Introduce SLOG in KEB


PK85:
The following [comment](https://github.com/kyma-project/kyma-environment-broker/issues/497#issuecomment-2049161710) lists all places where we import different log implementations. Change them to use slog implementation.

- consider to remove logs from storage layer
- talk with SRE if they still have alerts based on some logs from KEB

KsaweryZietara:
SRE let us know that their alerts are based on standard and custom metrics. The expressions are built around them when evaluating.

PR #1531, opened 2024-12-05
Use slog in Account Cleanup Job

KsaweryZietara:
**Description**

- use `slog` instead of `logrus` in Account Cleanup Job.

**Related issue(s)**
See also #1469

### DIFF
--- a/internal/cis/client.go
+++ b/internal/cis/client.go
@@ -5,12 +5,12 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"log/slog"
 	"net/http"
 	"strconv"
 	"time"
 
 	kebError "github.com/kyma-project/kyma-environment-broker/internal/error"
-	"github.com/sirupsen/logrus"
 	"golang.org/x/oauth2/clientcredentials"
 )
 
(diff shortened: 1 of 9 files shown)


internal/cis/client.go
```diff
@@ -52,7 +51,6 @@ func NewClient(ctx context.Context, config Config, log logrus.FieldLogger) *Clie
 	return &Client{
 		httpClient: httpClientOAuth,
 		config:     config,
-		log:        log.WithField("client", "CIS-2.0"),
```
szwedm, line 55: Client version won't be visible in logs. This may cause trouble while debugging issues.

internal/cis/client_ver_1.go
```diff
@@ -39,7 +38,6 @@ func NewClientVer1(ctx context.Context, config Config, log logrus.FieldLogger) *
 	return &ClientVer1{
 		httpClient: httpClientOAuth,
 		config:     config,
-		log:        log.WithField("client", "CIS-1.0"),
```
szwedm, line 42: Client version won't be visible in logs. This may cause trouble while debugging issues.

PR #1550, opened 2024-12-09
Use slog in runtime and expiration endpoints

KsaweryZietara:
**Description**

- use `slog` instead of `logrus` in runtime and expiration endpoints.

**Related issue(s)**
See also #1469

### DIFF
--- a/internal/runtime/handler_test.go
+++ b/internal/runtime/handler_test.go
@@ -3,8 +3,10 @@ package runtime_test
 import (
 	"encoding/json"
 	"fmt"
+	"log/slog"
 	"net/http"
 	"net/http/httptest"
+	"os"
 	"testing"
 	"time"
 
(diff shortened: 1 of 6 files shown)


internal/runtime/handler_test.go
```diff
@@ -62,7 +63,11 @@ func TestRuntimeHandler(t *testing.T) {
 		err = instances.Insert(testInstance2)
 		require.NoError(t, err)
 
-		runtimeHandler := runtime.NewHandler(db, 2, "", provisionerClient, k8sClient, kimConfig, logrus.New())
+		log := slog.New(slog.NewTextHandler(os.Stdout, &slog.HandlerOptions{
```
jaroslaw-pieszka, line 66: Do we need this to be duplicated so many times?

jaroslaw-pieszka (approved):

PR #1564, opened 2024-12-11
Use slog in queues

KsaweryZietara:
**Description**

- use `slog` instead of `logrus` in queues and orchestration steps.

**Related issue(s)**
See also #1469


cmd/broker/suite_test.go
```diff
@@ -107,9 +106,6 @@ func NewOrchestrationSuite(t *testing.T, additionalKymaVersions []string) *Orche
 		Level: slog.LevelInfo,
 	}))
 
-	logs := logrus.New()
-	logs.Formatter.(*logrus.TextFormatter).TimestampFormat = "15:04:05.000"
```
jaroslaw-pieszka, line 111: How timestamps are formatted when using slog?

jaroslaw-pieszka (approved):

PR #1604, opened 2025-01-02
Logging Worker Execution Times Only if Limit Exceeded

ralikio:
Removing execution time for given worker in removeTimeIfInWarnMargin function to track only workers that exceed allowed margin time. Added additional workerLastKeys map to keep worker keys for logging in worker summary go routine.

**Description**

* Removing execution time for given worker in removeTimeIfInWarnMargin function to track only workers that exceed allowed margin time.
* Added additional workerLastKeys map to keep worker keys for logging in worker summary go routine.

**Related issue(s)**

#1473

### DIFF
--- a/internal/process/queue.go
+++ b/internal/process/queue.go
@@ -1,7 +1,6 @@
 package process
 
 import (
-	"bytes"
 	"fmt"
 	"log/slog"
 	"runtime/debug"
--- a/internal/process/queue.go
+++ b/internal/process/queue.go
@@ -115,6 +115,7 @@ func (q *Queue) worker(queue workqueue.RateLimitingInterface, process func(key s
 
 				defer func() {
 					if err := recover(); err != nil {
+						q.removeTimeIfInWarnMargin(id, workerNameId, log)
 						log.Error(fmt.Sprintf("panic error from process: %v. Stacktrace: %s", err, debug.Stack()))
 					}
 					queue.Done(key)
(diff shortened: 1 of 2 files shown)


internal/process/queue.go
```diff
@@ -139,35 +143,43 @@ func (q *Queue) worker(queue workqueue.RateLimitingInterface, process func(key s
...
+	processingStart, ok := q.workerExecutionTimes[workerNameId]
+
+	if !ok {
+		log.Warn(fmt.Sprintf("worker %s has no start time", workerNameId))
+		return
+	}
+
+	processingTime := time.Since(processingStart)
+
+	if processingTime > q.warnAfterTime {
```
jaroslaw-pieszka, line 163: Will we log here information when the worker is idle?
In case there is nothing in the queue to process for a long time so the last execution time (in `q.workerExecutionTimes[workerNameId]` is very old.

ralikio: In the case the time from `workerExecutionTimes` is removed in https://github.com/kyma-project/kyma-environment-broker/pull/1604/files#diff-e94ea6ee51e7a52f24c56388c5828851ee956a6cb5d30cf83a6a82cef2f7784cR119 before a chance to log it hence not producing idle logs.

ralikio: Iddle logs were produced in the healthcheck go routine.

piotrmiskiewicz (approved):

Issue #1588, opened 2024-12-18
e2e tests refactor to go language


PK85:
**AC**

- [x] 1) refactor and unify e2e tests to go language. No js code anymore.
  - [x] 1.1) main goal is to keep test code simple and transparent. They must be easy to read.  
- [x] 2) introduce new tests, do not remove JS ones till they will work for longer time

Not mandatory, but you can take into account:

- skr tests with teraform
https://github.com/orgs/kyma-project/projects/38/views/1?filterQuery=test&pane=issue&itemId=90822016 

- gh actions better step overview
https://github.com/orgs/kyma-project/projects/38/views/1?filterQuery=test&pane=issue&itemId=53512049

- retriger all failed
https://github.com/orgs/kyma-project/projects/38/views/1?filterQuery=test&pane=issue&itemId=57061787

MarekMichali:
Keep in mind that `retriger all failed` can block most of the workers for 1 hour and other teams won't be able to perform their actions. This functionality was intentionally not implemented because, if I remember correctly, we did not have that many workers available at the time. Perhaps this has changed and there is no longer a risk of this happening.

MarekMichali:
- [ ] use slog not logrus

MarekMichali:
Issue to switch to new solution https://github.com/kyma-project/kyma-environment-broker/issues/1681

PR #1605, opened 2025-01-03
SKR tester: update machineType

MarekMichali:
**Description**

- Add command that updates the machineType

**Related issue(s)**

See #1588

### DIFF
--- a/testing/e2e/skr-tester/pkg/broker/client.go
+++ b/testing/e2e/skr-tester/pkg/broker/client.go
@@ -246,23 +246,13 @@ func (c *BrokerClient) ProvisionInstance(instanceID, planID, region string, btpO
 	return c.CallBroker(payload, endpoint, "PUT")
 }
 
-func (c *BrokerClient) UpdateInstance(instanceID string, customParams, btpOperatorCreds map[string]interface{}, isMigration bool) (map[string]interface{}, error) {
+func (c *BrokerClient) UpdateInstance(instanceID string, customParams map[string]interface{}) (map[string]interface{}, error) {
 	payload := map[string]interface{}{
 		"service_id": kymaServiceID,
 		"context": map[string]interface{}{
 			"globalaccount_id": c.GlobalAccountID,
-			"isMigration":      isMigration,
 		},
 		"parameters": customParams,
-	}
-
-	if btpOperatorCreds != nil {
-		payload["context"].(map[string]interface{})["sm_operator_credentials"] = map[string]interface{}{
-			"clientid":     btpOperatorCreds["clientid"],
-			"clientsecret": btpOperatorCreds["clientsecret"],
-			"sm_url":       btpOperatorCreds["smURL"],
-			"url":          btpOperatorCreds["url"],
-		}
 	}
 
 	endpoint := fmt.Sprintf("service_instances/%s?accepts_incomplete=true", instanceID)
--- a/testing/e2e/skr-tester/pkg/broker/client.go
+++ b/testing/e2e/skr-tester/pkg/broker/client.go
@@ -14,7 +14,7 @@ import (
 
 const (
 	scope                    = "broker:write"
-	kymaServiceID            = "47c9dcbf-ff30-448e-ab36-d3bad66ba281"
+	KymaServiceID            = "47c9dcbf-ff30-448e-ab36-d3bad66ba281"
 	trialPlanID              = "7d55d31d-35ae-4438-bf13-6ffdfa107d9f"
 	defaultExpirationSeconds = 600
 )
@@ -212,7 +212,7 @@ func (c *BrokerClient) GetCatalog() (map[string]interface{}, error) {
 
 func (c *BrokerClient) BuildPayload(name, instanceID, planID, region string, btpOperatorCreds map[string]interface{}) map[string]interface{} {
 	payload := map[string]interface{}{
-		"service_id": kymaServiceID,
+		"service_id": KymaServiceID,
 		"plan_id":    planID,
 		"context": map[string]interface{}{
 			"globalaccount_id": c.GlobalAccountID,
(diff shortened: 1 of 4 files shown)


testing/e2e/skr-tester/pkg/broker/client.go
```diff
@@ -246,25 +246,15 @@ func (c *BrokerClient) ProvisionInstance(instanceID, planID, region string, btpO
...
 	payload := map[string]interface{}{
 		"service_id": kymaServiceID,
 		"context": map[string]interface{}{
 			"globalaccount_id": c.GlobalAccountID,
-			"isMigration":      isMigration,
 		},
 		"parameters": customParams,
 	}
 
-	if btpOperatorCreds != nil {
```
szwedm, line 259: Is update of SAP BTP service operator credentials no longer supported?

MarekMichali: Not at the moment. I will add it when needed.

MarekMichali: I want to keep it clean and add functionality that is needed. Previously it was added more as a placeholder as the function was rewritten from nodejs.

testing/e2e/skr-tester/pkg/command/update.go
```diff
@@ -0,0 +1,194 @@
...
+		return errors.New("you must specify the planID and instanceID")
+	}
+}
+
+func getCurrentMachineType(instanceID string) (*string, error) {
+	_, err := exec.Command("kcp", "login").Output()
+	if err != nil {
+		return nil, fmt.Errorf("failed to run kcp login: %v", err)
+	}
+	output, err := exec.Command("kcp", "rt", "-i", instanceID, "-ojson", "--runtime-config").Output()
```
szwedm, line 117: Try the following command: `kcp rt -i <INSTANCE_ID> --runtime-config -o custom=":{.runtimeConfig.spec.shoot.provider.workers[0].machine.type}"`
You should get a response with a empty line (empty header) and a machine type.

MarekMichali: Tried it but the `workers[0]` is not working

testing/e2e/skr-tester/pkg/broker/client.go
szwedm: Ok, let's focus on machine type update in this PR. I think we will need the possibility to update the credentials in the context for one of the e2e tests, but we can add it later.

testing/e2e/skr-tester/pkg/command/update.go
MarekMichali: Changed.

testing/e2e/skr-tester/pkg/command/update.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+func NewUpdateCommand() *cobra.Command {
+	cmd := UpdateCommand{}
+	cobraCmd := &cobra.Command{
+		Use:     "update",
+		Aliases: []string{"u"},
+		Short:   "Update the instnace",
+		Long:    "Update the instnace",
+		Example: "skr-tester update -i instanceID --updateMachineType                            Update the instance with new machineType.",
```
szwedm, line 31: I think it's worth adding the required `planID` param to the example.

MarekMichali: Added.

testing/e2e/skr-tester/pkg/command/update.go
```diff
@@ -0,0 +1,157 @@
...
+		if !ok {
+			return errors.New("service is not a map[string]interface{}")
+		}
+		if cmd.updateMachineType {
+			currentMachineType, err := getCurrentMachineType(cmd.instanceID)
+			if err != nil {
+				return fmt.Errorf("failed to get current machine type: %v", err)
+			}
+			fmt.Printf("Current machine type: %s\n", *currentMachineType)
+			if serviceMap["id"] != "47c9dcbf-ff30-448e-ab36-d3bad66ba281" {
```
szwedm, line 67: What do you think about checking the service as a first condition for the code execution? We are only interested in updating the Kyma instance, so we could get the service by ID from `services` slice and work with that instead iterating through the slice. You can add a const to hold `47c9dcbf-ff30-448e-ab36-d3bad66ba281` value. Looks like the ID is interface{} type when you read it from the map (type of map[string]interface{}).

MarekMichali: Moved the condition to the top of the services loop, and now I use the const form broker package with the serviceID.

testing/e2e/skr-tester/pkg/command/update.go
```diff
@@ -0,0 +1,157 @@
...
+			plans, ok := serviceMap["plans"].([]interface{})
+			if !ok {
+				return errors.New("plans field not found or invalid in serviceMap")
+			}
+			for _, p := range plans {
+				planMap, ok := p.(map[string]interface{})
+				if !ok || planMap["id"] != cmd.planID {
+					continue
+				}
+				updateParams, err := extractUpdateParams(planMap)
```
szwedm, line 79: nit: I find `updateParams` name a little confusing. `extractUpdateParams` returns allowed machine type values.

testing/e2e/skr-tester/pkg/command/update.go
```diff
@@ -0,0 +1,157 @@
...
+				if err != nil {
+					return fmt.Errorf("failed to extract update parameters: %v", err)
+				}
+				if len(updateParams) < 2 {
+					continue
+				}
+				for i, m := range updateParams {
+					if m == *currentMachineType {
+						newMachineType := updateParams[(i+1)%len(updateParams)].(string)
```
szwedm: This looks nice for a flow when the command is run without providing any machine type as a parameter to the command. I think it may be worthwhile to implement the flow when machine type name is provided as a parameter.

testing/e2e/skr-tester/pkg/command/update.go
szwedm: Nice 👍

testing/e2e/skr-tester/pkg/command/update.go
MarekMichali: I will add your idea to the ticket as an AC if you agree to wait with this feature. I would like to get the tests running as they are now, so we can get rid of nodejs.

szwedm: Sure, we can add extra functionality in next PRs.

MarekMichali: https://github.com/kyma-project/kyma-environment-broker/issues/1588#issuecomment-2565442026

szwedm (approved):
/lgtm

Issue #1520, opened 2024-12-03
Configurable HAP mode per plan


ebensom:
**Description**

Currently it is hardcoded in KEB provisioning process whether to use "shared" or "dedicated" hyperscaler account pool (HAP) via [account provider](https://github.com/kyma-project/kyma-environment-broker/blob/main/common/hyperscaler/account_provider.go). For trial and sap-converged-cloud plans, shared pool, while for otherts, the "dedicated" account pool is used.

The pool mode (i.e. "shared" or "dedicated") shall be made configurable per plan, or otherwise it shall be made configurable which plans should use shared HAP.
 
**Reasons**

Kyma runtime Sovereign Cloud / NS2 would work using shared pools for all supported plans.

**Attachments**

PK85:
Additional ACs:
-  it could not be only per plan, we need to support that even per region. Like cf-sa30 or eu regions, they use shared pools too.

PK85:
Result of my detailed analysis:
- plans **trial (hyperscaler-type: trial)** and **sap-converged-cloud (hyperscaler-type: openstack_eu-de-1)** use shared subscriptions that are marked with **shared** lag
- eu_access **aws cf-eu11** and **azure cf-ch20** use dedicated subscription that are marked with **eu_access** flag
- assured workloads hyperscaler-type set to gcp_cf-sa30 use dedicated subscription

ralikio:
The example no 2 has been dismissed because shared pools do not use euAccess functionality.

To address the above comments and cover all cases (especially about removing hardcoded configuration) I have analyzed the existing pools thinking how we can implement it starting from "zero".

Let's imaging that HAP pool functionality does not exist yet and we are doing selection of pools based only on simple HAP type (without any regions). The environment is populated with SecretBindings with label 'hyperscaler-type' (a new label, the same name as existing hyperscaler-type just for convenience, name can be tailored later) forming a pool of dedicated bindings per hyperscaler type that the plan is provisioned on:

![image](https://github.com/user-attachments/assets/484b1c8c-7295-40b7-9018-55b31c4dd363)

In this case KEB selects a binding based on region that is a physical attribute known during the provisioning process. This might be simple label construction process without any complicated implementation.

[...]

PR #1609, opened 2025-01-06
Configurable HAP Shared Secrets Plans and Regions

ralikio:
**Description**

The following PR implements #1520 feature request. It introduces two configuration lists that allows for specifying which plans or platform regions should be provisioned with shared gardener secrets.

- structures for "whitelisting" plans and regions that should be provisioned with shared secrets,
- default configuration values,
- usage of above in resolve credentials step,
- docs update.


PK85:
see: https://sap-btp.slack.com/archives/G01L562297B/p1736235785889099

PK85:
I did a code review, and decided to left more high level review, please see: https://github.com/kyma-project/kyma-environment-broker/issues/1520

We will talk about that tomorrow

internal/utils/whitelist.go
```diff
@@ -0,0 +1,39 @@
...
+func (t Whitelist) String() string {
+	keys := make([]string, 0, len(t))
+	for item := range t {
+		keys = append(keys, item)
+	}
+
+	sort.Strings(keys)
+
+	output := ""
+	for _, item := range keys {
```
jaroslaw-pieszka, line 34: Why strings.Join is not appropriate?

internal/utils/whitelist.go
```diff
@@ -0,0 +1,39 @@
+package utils
+
+import (
+	"sort"
+	"strings"
+)
+
+type Whitelist map[string]struct{}
```
jaroslaw-pieszka, line 8: We define some date structure (even kind of ADT) but we use the name not related to its behavior.  There is nothing about `whitelisting`, just set of strings with contains operation.

docs/contributor/03-10-hyperscaler-account-pool.md
```diff
@@ -82,3 +83,35 @@ metadata:
     tenant-name: {TENANT_NAME}
     hyperscaler-type: "gcp_cf-sa30"
 ```
+
+## HAP Configuration
+
+<!-- selection algorithm -->
+When selecting a Secret for a given hyperscaler, KEB evaluates a set of rules to determine what secret bindings it should query for a given plan and region that SKR is provisioned in. The search looks for a secret bindings with `hyperscaler-type` label only (always used in the search) or additionally with one of or both `euAccess: true` and `shared: true` labels. 
+
+`hyperscaler-type` label is the one that does contain region information in the format `hyperscaler_type: <HYPERSCALER_NAME>[_<PLATFORM_REGION>][_<CLUSTER_REGION>]`, where both `_<PLATFORM_REGION>` and `_<CLUSTER_REGION>` are optional.
```
jaroslaw-pieszka, line 92: Could we stick to `hyperscaler region` (HYPERSCALER_REGION accordingly) as in #1520

docs/contributor/03-10-hyperscaler-account-pool.md
```diff
@@ -82,3 +83,35 @@ metadata:
...
+* `hap.sharedRules` - if evaluated to true the `shared: true` label is used when searching, refered to as shared based search,
+* `hap.euAccessRules` - if evaluated to true the `shared: true` label is used when searching, refered to as euAccess based search.
+
+![alt text](image.png)
+
+<!-- rules format -->
+Each configuration consists of a list (The list entries are separated with semicolons) of plans that the rule applies to. Additionally, a plan can be extended with `:region` suffix that makes the rule evaluate to true only if a cluster is provisioned in the specified region.
+
+The property accepts a list of key value in the format `<PLAN_ID_1>:<REGION_ID_1>;<PLAN_ID_2>:<REGION_ID_2>`. Either plan or region (but never both) can be specified as wildcard `*` meaning all plans or regions should apply for specific second value. The following example lists valid and invalid configuration values:
+* `trial` - valid, rule will be evaluated to true for the `trial` plan in all regions,
```
jaroslaw-pieszka, line 106: How do we deal with trial in case trial provider is changed? What I mean is that we possibly need not plan only, but provider to determine labels.

docs/contributor/03-10-hyperscaler-account-pool.md
ralikio: I want something shorter. HAP_REGION?

docs/contributor/03-10-hyperscaler-account-pool.md
ralikio: If you have one provision request for plan trial and provider azure and second one for trial aws they would never use the same pool because, by default, we always select a pool by the provider/hyperscaler type. In case provider changes the new provider must have correct pools been set up beforehand.

ralikio:
POC. I am dividing the PR into smaller chunks.

Issue #952, opened 2024-07-25
Extend Kyma Provisioning Parameters to allow multiple Worker Node Pools definitions


PK85:
See:

https://github.com/kyma-project/kyma/issues/18709#issuecomment-2251353514

Plan:
1) We keep our main worker pool nodes as they are. It will be for Kyma workloads
2) We need to support additionalWorkerNodes for customers (we decided that they are by default min 3 and three zones)

**AC**
- [x] 0) extend schema with `_load_current_config` parameter  (CIS released staging: DONE, canary: DONE, 
prod: DONE)
https://github.com/kyma-project/kyma-environment-broker/pull/1654
- [x] 1) design KEB schema API, validations, defaults. Present to the team.
- [x] 2) use our preview plan where new schema is visible, and we update RuntimeCR
https://github.com/kyma-project/kyma-environment-broker/pull/1659, https://github.com/kyma-project/kyma-environment-broker/pull/1668, 
https://github.com/kyma-project/kyma-environment-broker/pull/1671, https://github.com/kyma-project/kyma-environment-broker/pull/1691
- [x] 3) Check how frogs wants to implement that in KIM, see: https://github.com/kyma-project/infrastructure-manager/issues/46#issuecomment-2548046300
- [x] 4) Implementation for all plans. DO NOT expose additionalWorkerPool for trial and free. Azure_lite max node same as main worker pool and HA not available in the schema, always one zone. 
https://github.com/kyma-project/kyma-environment-broker/pull/1695, https://github.com/kyma-project/kyma-environment-broker/pull/1697
- [x] 5) new e2e SKR tests https://github.com/kyma-project/kyma-environment-broker/pull/1721, https://github.com/kyma-project/kyma-environment-broker/pull/1723
- [x] 6) docs https://github.com/kyma-project/kyma-environment-broker/pull/1617, https://github.com/kyma-project/kyma-environment-broker/pull/1678
    - [ ] 6.1 Help Portal documentation:
          - [x] [Available Plans in the Kyma Environment](https://help.sap.com/docs/btp/sap-business-technology-platform/available-plans-in-kyma-environment)
          - [ ] [Provisioning and Updating Parameters in the Kyma Environment](https://help.sap.com/docs/btp/sap-business-technology-platform/provisioning-and-update-parameters-in-kyma-environment)
          - [ ] RN
- [x] 7) deploy (featureFlag introduced, when deployed on prod we can remove FF)
  - [x] 7.1) dev
[...]

PR #1668, opened 2025-01-23
Introduce HA zones

KsaweryZietara:
**Description**

- introduce HA zones field to additional worker node pools schema,
- validate if names of additional worker node pools are unique,
- adjust tests.

**Related issue(s)**
See also #952

### DIFF
--- a/common/runtime/model.go
+++ b/common/runtime/model.go
@@ -406,10 +406,10 @@ func (a AdditionalWorkerNodePool) Validate() error {
 	return nil
 }
 
-func (a AdditionalWorkerNodePool) ValidateHAZones(instanceAdditionalWorkerNodePools []AdditionalWorkerNodePool) error {
-	for _, instanceAdditionalWorkerNodePool := range instanceAdditionalWorkerNodePools {
-		if a.Name == instanceAdditionalWorkerNodePool.Name {
-			if !a.HAZones && instanceAdditionalWorkerNodePool.HAZones {
+func (a AdditionalWorkerNodePool) ValidateDisablingHAZones(currentAdditionalWorkerNodePools []AdditionalWorkerNodePool) error {
+	for _, currentAdditionalWorkerNodePool := range currentAdditionalWorkerNodePools {
+		if a.Name == currentAdditionalWorkerNodePool.Name {
+			if !a.HAZones && currentAdditionalWorkerNodePool.HAZones {
 				return fmt.Errorf("HA zones cannot be disabled for %s additional worker node pool", a.Name)
 			}
 		}
--- a/common/runtime/model.go
+++ b/common/runtime/model.go
@@ -391,13 +391,28 @@ type ModuleDTO struct {
 type AdditionalWorkerNodePool struct {
 	Name          string `json:"name"`
 	MachineType   string `json:"machineType"`
+	HAZones       bool   `json:"haZones"`
 	AutoScalerMin int    `json:"autoScalerMin"`
 	AutoScalerMax int    `json:"autoScalerMax"`
 }
 
 func (a AdditionalWorkerNodePool) Validate() error {
 	if a.AutoScalerMin > a.AutoScalerMax {
-		return fmt.Errorf("AutoScalerMax %v should be larger than AutoScalerMin %v for %s worker node pool", a.AutoScalerMax, a.AutoScalerMin, a.Name)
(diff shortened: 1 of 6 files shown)


internal/broker/instance_create.go
```diff
@@ -412,6 +415,17 @@ func supportsAdditionalWorkerNodePools(planID string) bool {
 	return false
 }
 
+func areNamesUnique(pools []pkg.AdditionalWorkerNodePool) bool {
```
jaroslaw-pieszka, line 418: You can manage without bool. :)

1. Count
2. Make set i.e. map[string]struct{}
3. Insert elements
4. Count elements in the set
5. compare checks in 1. and 4.

jaroslaw-pieszka: However your solution is faster since we do not need to process all elements (on average).

common/runtime/model.go
```diff
@@ -391,13 +391,28 @@ type ModuleDTO struct {
...
-		return fmt.Errorf("AutoScalerMax %v should be larger than AutoScalerMin %v for %s worker node pool", a.AutoScalerMax, a.AutoScalerMin, a.Name)
+		return fmt.Errorf("AutoScalerMax %v should be larger than AutoScalerMin %v for %s additional worker node pool", a.AutoScalerMax, a.AutoScalerMin, a.Name)
+	}
+	if a.HAZones && a.AutoScalerMin < 3 {
+		return fmt.Errorf("AutoScalerMin %v should be at least 3 when HA zones are enabled for %s additional worker node pool", a.AutoScalerMin, a.Name)
+	}
+	return nil
+}
+
+func (a AdditionalWorkerNodePool) ValidateHAZones(instanceAdditionalWorkerNodePools []AdditionalWorkerNodePool) error {
```
jaroslaw-pieszka, line 409: It is not clear at first glance what `validate` means.

internal/broker/instance_update_test.go
```diff
@@ -679,6 +687,49 @@ func TestUpdateAdditionalWorkerNodePools(t *testing.T) {
...
+		ServiceID:       "",
+		PlanID:          PreviewPlanID,
+		RawParameters:   json.RawMessage(`{"additionalWorkerNodePools": [{"name": "name-1", "machineType": "m6i.large", "haZones": false, "autoScalerMin": 3, "autoScalerMax": 20}]}`),
+		PreviousValues:  domain.PreviousValues{},
+		RawContext:      json.RawMessage("{\"globalaccount_id\":\"globalaccount_id_1\", \"active\":true}"),
+		MaintenanceInfo: nil,
+	}, true)
+
+	// then
+	assert.EqualError(t, err, "HA zones cannot be disabled for name-1 additional worker node pool")
```
jaroslaw-pieszka, line 730: So we test the scenario with not allowed transition from haZones true->false. Am I right?

internal/broker/instance_create.go
jaroslaw-pieszka, line 418: Do you think this is so elementary that we do not need unit test for it?

common/runtime/model.go
jaroslaw-pieszka: Are there any test cases for this function?

internal/broker/instance_create.go
jaroslaw-pieszka: You could consider golang idiom (disputable) and use set (map[string]struct{}{})

common/runtime/model.go
KsaweryZietara: Will `ValidateDisablingHAZones` and `currentAdditionalWorkerNodePools` names be more accurate? I'm open to suggestions regarding naming.

I will add this test scenario (false -> true).

internal/broker/instance_update_test.go
KsaweryZietara: Yes, is is not possible to disable HA.

internal/broker/instance_create.go
KsaweryZietara: I will use the set and add unit tests.

common/runtime/model.go
jaroslaw-pieszka: > Once HA for some worker node pools is enabled in provisioning or update, it cannot be disabled in next update operation.
> 
> Will `ValidateDisablingHAZones` and `currentAdditionalWorkerNodePools` names be more accurate? I'm open to suggestions regarding naming.
> 
> I will add this test scenario (false -> true).

Would be perfect.

jaroslaw-pieszka (approved):

PR #1691, opened 2025-01-29
Additional workers in RuntimeCR

KsaweryZietara:
**Description**

- bump `infrastructure-manager` dependency,
- configure additional workers in RuntimeCR during provisioning and updates.

**Related issue(s)**
See also #952


cmd/broker/broker_suite_test.go
```diff
@@ -1276,3 +1278,18 @@ func createResource(t *testing.T, gvk schema.GroupVersionKind, k8sClient client.
 	err := k8sClient.Create(context.TODO(), object)
 	assert.NoError(t, err)
 }
+
+func (s *BrokerSuiteTest) assertAdditionalWorkerIsCreated(t *testing.T, provider imv1.Provider, name, machineType string, autoScalerMin, autoScalerMax, zonesNumer int) {
```
jaroslaw-pieszka, line 1282: So we search if there is specific additional worker pool specified by name.
And if we what two assert e.g. 3 additional worker  pools we search three times the array (O(n)).
But this is nit-picking. Performance will not be a problem.

internal/process/provisioning/create_runtime_resource_step.go
```diff
@@ -250,6 +251,49 @@ func (s *CreateRuntimeResourceStep) createShootProvider(operation *internal.Oper
...
 		}
 	}
+
+	if len(operation.ProvisioningParameters.Parameters.AdditionalWorkerNodePools) > 0 {
+		additionalWorkerNodePoolsMaxUnavailable := intstr.FromInt32(int32(0))
+		workers := make([]gardener.Worker, 0, len(operation.ProvisioningParameters.Parameters.AdditionalWorkerNodePools))
+
+		for _, additionalWorkerNodePool := range operation.ProvisioningParameters.Parameters.AdditionalWorkerNodePools {
+			workerZones := values.Zones
+			if !additionalWorkerNodePool.HAZones {
```
jaroslaw-pieszka, line 261: So if there is no HA, we pick up one zone randomly from previously randomly assigned zones for the cluster.
Am I correct?

So far we used random.shuffle. Please consider this approach as "conventional":
```
	rand.Shuffle(len(zones), func(i, j int) { zones[i], zones[j] = zones[j], zones[i] })
	return zones[:zonesCount]
```

internal/process/update/update_runtime_step.go
```diff
@@ -59,6 +73,54 @@ func (s *UpdateRuntimeStep) Run(operation internal.Operation, log *slog.Logger)
...
+		values, err := provider.GetPlanSpecificValues(&operation, s.config.MultiZoneCluster, s.config.DefaultTrialProvider, s.useSmallerMachineTypes, s.trialPlatformRegionMapping,
+			s.config.DefaultGardenerShootPurpose, s.config.ControlPlaneFailureTolerance)
+		if err != nil {
+			return s.operationManager.OperationFailed(operation, fmt.Sprintf("while calculating plan specific values: %s", err), err, log)
+		}
+
+		additionalWorkerNodePoolsMaxUnavailable := intstr.FromInt32(int32(0))
+		workers := make([]gardener.Worker, 0, len(operation.UpdatingParameters.AdditionalWorkerNodePools))
+
+		for _, additionalWorkerNodePool := range operation.UpdatingParameters.AdditionalWorkerNodePools {
```
jaroslaw-pieszka, line 86: Is there any way to avoid code duplication?
This is the same code (at first glance) that in create runtime step.

cmd/broker/broker_suite_test.go
KsaweryZietara: I could use a `map[string]*v1beta1.Worker` to store workers by name, reducing lookup time from O(n) to O(1). However, since the number of additional worker node pools in tests is always fewer than three, this optimization seems unnecessary.

internal/process/update/update_runtime_step.go
KsaweryZietara: I've extracted the duplicated logic into a `CreateAdditionalWorkers` function.

jaroslaw-pieszka (approved):

Issue #1587, opened 2024-12-18
Replace gorilla/mux with standard library


PK85:
Continuation of the task to investigate possible replacements for the current router in KEB to achieve compatibility with the latest version of [brokerapi](https://github.com/pivotal-cf/brokerapi). Replacing gorillxa/mux with standard library should ensure full compatibility and be future-proof.

See: https://github.com/kyma-project/kyma-environment-broker/issues/1587#issuecomment-2559680141

**AC**

- [x] 1) if everything works then we go with that solution and remove fork and use upstream brokerapi repo

MarekMichali:
We will wait for https://github.com/pivotal-cf/brokerapi/pull/340 to be released. Then we can use this https://github.com/kyma-project/kyma-environment-broker/pull/1525

MarekMichali:
We must wait for PK to remove the brokerAPI fork.

PR #1694, opened 2025-01-30
Replace gorilla/mux with router from std lib

MarekMichali:
**Description**

- Replace gorilla/mux with router from std lib
- Use latest version of brokerAPI

**Related issue(s)**

See #1587


internal/middleware/region.go
```diff
@@ -16,12 +14,11 @@ const (
 	requestRegionKey key = iota + 1
 )
 
-func AddRegionToContext(defaultRegion string) mux.MiddlewareFunc {
+func AddRegionToContext(defaultRegion string) MiddlewareFunc {
```
szwedm, line 17: Somehow this middleware is called twice before the request reaches dedicated endpoint.

internal/httputil/router.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+}
+
+func (r *Router) Handle(pattern string, handler http.Handler) {
+	for i := len(r.middlewares) - 1; i >= 0; i-- {
+		handler = r.middlewares[i](handler)
+	}
+	r.ServeMux.Handle(pattern, handler)
+}
+func (r *Router) HandleWithoutMiddleware(pattern string, handler http.Handler) {
```
szwedm, line 36: Method is not used anywhere.

internal/httputil/response_recorder.go
```diff
@@ -0,0 +1,25 @@
+package httputil
+
+import "net/http"
+
+type ResponseRecorder struct {
```
szwedm, line 5: Decorator may be redundant. Check Header() method of http.ResponseWriter interface.

internal/middleware/region.go
MarekMichali: Fixed. The problem was that the middleware set for the parent router was also called for the child router. I moved the middleware definition to the child router since it is not used by any endpoints from the parent router.

internal/httputil/router.go
MarekMichali: Removed.

internal/httputil/response_recorder.go
MarekMichali: I didn't find anything in the Header()

szwedm (approved):
/lgtm

internal/httputil/response_recorder.go
szwedm: Ok, thanks for checking.

szwedm (approved):
/lgtm

Issue #1698, opened 2025-01-30
Rule Entry Parser


ralikio:
**Description**

Implement a rule parser in accordance to [Rule Format](https://github.com/kyma-project/kyma-environment-broker/pull/1675/files#diff-124a86ccf11730da9fb89284c51142f5aa3cd46c816bc9a8ddad8f30bb436b2bR35). 
- [x] Parse in accordance to defined format.
- [x] Expect only attributes described in `Rule Attributes` section. 
- [x] Consider using ANTLR or other parser/lexer: https://blog.gopheracademy.com/advent-2017/parsing-with-antlr4-and-go/
- [x] Consider using regex validation and reading expected rule entry parts with simple strings operations or string splitting

**Attachments**

#1520 #1675 

Example lexer:
```
// DELETE THIS CONTENT IF YOU PUT COMBINED GRAMMAR IN Parser TAB
lexer grammar ExprLexer;

EQ : '=' ;
COMMA : ',' ;
LPAREN : '('  ;
RPAREN : ')' ;
ARROW : '->' ;
ASTERIX : '*' ;

WS: [ \t\n\r\f]+ -> skip ;

// TODO: fill in missing plans
PLAN: 'azure'
    | 'azure_lite'
    | 'aws'
    | 'gcp'
    | 'trial'
    | 'free'
    | 'sap-converged-cloud'
    ;
    
INPUT_ATTR: 'PR'
    | 'HR'
    ;
    
OUTPUT_ATTR: 'S'
    | 'EU'
    ;
    
VAL: [a-zA-Z_][a-zA-Z_0-9-]*;
```

Example parser
```
grammar ExprParser;
options { tokenVocab=ExprLexer; }

ruleEntry
    : entry EOF
    ;

entry: PLAN
    | PLAN inputAttrInParen
    | PLAN inputAttrInParen ARROW outputAttrList
    ;

inputAttrInParen: LPAREN RPAREN
    | LPAREN inputAttrList RPAREN
    ;
    
outputAttrInParen: LPAREN RPAREN
    | LPAREN outputAttrList RPAREN
    ;
    
inputAttrList: inputAttrVal
    | inputAttrVal COMMA inputAttrList
    ;

outputAttrList: outputAttrVal
    | outputAttrVal COMMA outputAttrList
    ;
    
inputAttrVal: INPUT_ATTR EQ VAL
    | INPUT_ATTR EQ ASTERIX
    ;

[...]

PR #1703, opened 2025-01-31
Expression Parsing with Antlr and Simple String Splitting

ralikio:
**Description**
Comparison between Antrl parser and simple string splitting to implement #1520.

**Related issue(s)**

[1698](https://github.com/kyma-project/kyma-environment-broker/issues/1698)


PK85:
General comment about hap CLI version 0.0.5"
- 1) Looks that generally parsing rules works, but found below issues:
  - 1.1) Search labels are not displayed
  - 1.2) Some errors for rules could be more specific, like: 
    - 1.2.1) If wrong input attr, output attr, plan then suggest users what are correct possible values
    - 1.2.2) Position of the incorrect value is not specified

cmd/parser/parse.go
```diff
@@ -0,0 +1,356 @@
...
+
+	if cmd.noColor {
+		colorError = ""
+		colorOk = ""
+		colorNeutral = ""
+		colorMatched = ""
+	}
+
+	if cmd.match != "" && (!cmd.sort || !cmd.unique) {
+		cmd.cobraCmd.Printf("\tMatching is only supported when both priority and uniqnuess flags are specified.\n")
```
jaroslaw-pieszka, line 103:
```suggestion
		cmd.cobraCmd.Printf("\tMatching is only supported when both priority and uniqueness flags are specified.\n")
```

cmd/parser/parse.go
```diff
@@ -0,0 +1,356 @@
...
+	}
+
+	if cmd.sort {
+		allResults = rules.SortRuleEntries(allResults)
+		okResults = rules.SortRuleEntries(okResults)
+		errorResults = rules.SortRuleEntries(errorResults)
+	}
+
+	if cmd.unique {
+		uniqnuessSet := make(map[string]rules.ParsingResult)
```
jaroslaw-pieszka, line 151:
```suggestion
		uniquenessSet := make(map[string]rules.ParsingResult)
```

cmd/parser/parse.go
```diff
@@ -0,0 +1,356 @@
...
+			if result.Err != nil {
+				uniqueResults = append(uniqueResults, result)
+				continue
+			}
+
+			negativeSignatureKey := result.Rule.Plan
+			signatureKey := result.Rule.Plan
+			key := result.Rule.Plan
+			
+			key += "PR:"
```
jaroslaw-pieszka, line 169: Could we extract this literal as const?

cmd/parser/parse.go
```diff
@@ -0,0 +1,356 @@
...
+
+			negativeSignatureKey := result.Rule.Plan
+			signatureKey := result.Rule.Plan
+			key := result.Rule.Plan
+			
+			key += "PR:" 
+			signatureKey += "PR:"
+			negativeSignatureKey += "PR:"
+			if result.Rule.PlatformRegion == "" || result.Rule.PlatformRegion == "*" {
+				key += "*"
```
jaroslaw-pieszka, line 173: Consider extracting all literals used commonly across the code as constants.

cmd/parser/parse.go
```diff
@@ -0,0 +1,356 @@
...
+				key += result.Rule.PlatformRegion
+			
+				signatureKey += "attr"
+				negativeSignatureKey += "*"
+			}
+			
+			key += "HR:"
+			signatureKey += "HR:"
+			negativeSignatureKey += "HR:"
+			if result.Rule.HyperscalerRegion == "" || result.Rule.HyperscalerRegion == "*" {
```
jaroslaw-pieszka, line 190: Code duplication. We could possibly extract part of it as function and call it twice.

cmd/parser/parse.go
```diff
@@ -0,0 +1,356 @@
...
+			// if negativeSignatureExists && containsWildcards && cmd.signature {
+			if negativeSignatureExists && cmd.signature {
+				
+				resolvingSignaturePossibleRule := result.Rule.Combine(*negativeSignatureItem.Rule)
+
+			resolvingKey := resolvingSignaturePossibleRule.StringNoLabels()
+			_, resolvingRuleExists := resolvedRules[resolvingKey]
+
+				if !resolvingRuleExists {
+					err := fmt.Errorf("Duplicated negative signature with previously defined rule: '%s', consider introducing a resolving rule '%s'", negativeSignatureItem.Rule.StringNoLabels(), resolvingKey)
```
jaroslaw-pieszka, line 219: Negative signature is kind of cryptic for the user. We need to invent something simpler or add a section to documentation to explain it properly.

cmd/parser/parse.go
```diff
@@ -0,0 +1,356 @@
...
+					errorResults = append(errorResults, rules.ParsingResult{OriginalRule: result.OriginalRule, Err: err})
+
+					uniqueResults = append(uniqueResults, rules.ParsingResult{OriginalRule: result.OriginalRule, Err: err})
+					continue
+				}
+			}
+
+			alreadyExists := false
+			var item rules.ParsingResult
+			item, alreadyExists = uniqnuessSet[key]
```
jaroslaw-pieszka, line 230:
```suggestion
			item, alreadyExists = uniquenessSet[key]
```

cmd/parser/parse.go
```diff
@@ -0,0 +1,356 @@
...
+	
+			if !alreadyExists {
+
+				uniqnuessSet[key] = result
+				signatureSet[signatureKey] = result
+				uniqueResults = append(uniqueResults, result)
+
+			} else {
+				err := fmt.Errorf("Duplicated rule with previously defined rule: '%s'", item.Rule.StringNoLabels())
```
jaroslaw-pieszka, line 239: Wouldn't be better to use
```suggestion
				err := fmt.Errorf("Rule duplicates previously defined rule: '%s'", item.Rule.StringNoLabels())
```

cmd/parser/parse.go
```diff
@@ -0,0 +1,356 @@
...
+
+			cmd.cobraCmd.Printf("%s Error %s", colorError, colorNeutral)
+
+		} else {
+
+			if (cmd.match != "" && testDataForMatching != nil) {
+				matched := result.Rule.Matched(testDataForMatching)
+
+				if matched {
+					cmd.cobraCmd.Printf("%s Matched %s ", colorMatched, colorNeutral)
```
jaroslaw-pieszka, line 274: Could be extracted as function printing messages in given color and reverting neutral one.

cmd/parser/parse.go
```diff
@@ -0,0 +1,356 @@
...
+
+		if result.Err != nil {
+			cmd.cobraCmd.Printf(" %s", result.OriginalRule)
+			cmd.cobraCmd.Printf(" - %s", result.Err)
+		}
+
+		cmd.cobraCmd.Printf("\n")
+	}
+
+	if len(errorResults) != 0 {
```
jaroslaw-pieszka, line 293: So the function result is not correlated with length  of `errorResults`. 
We just return nil after printing the message.

cmd/parser/parse.go
```diff
@@ -0,0 +1,356 @@
...
+		cmd.cobraCmd.Printf("There are errors in your rule configuration. Fix above errors in your rule configuration and try again.\n")
```
jaroslaw-pieszka, line 294: Wouldn't the number of errors be convenient for the user?

cmd/parser/parse.go
```diff
@@ -0,0 +1,356 @@
...
+		return nil
+	}
+
+	return nil
+}
+
+func resolvingSignatureFormat(item rules.ParsingResult) string {
+	positiveSignature := item.Rule.Plan
+	if item.Rule.PlatformRegion == "*" || item.Rule.PlatformRegion != "" {
```
jaroslaw-pieszka, line 303:
```suggestion
	if item.Rule.PlatformRegion != "" {
```

cmd/parser/parse.go
```diff
@@ -0,0 +1,356 @@
...
+		positiveSignature += "PR:attr"
+	}
+
+	if item.Rule.HyperscalerRegion == "*" || item.Rule.HyperscalerRegion != "" {
```
jaroslaw-pieszka, line 307:
```suggestion
	if item.Rule.HyperscalerRegion != "" {
```

cmd/parser/parse.go
```diff
@@ -0,0 +1,356 @@
...
+	cmd.parser = &rules.SimpleParser{}
+
+	if cmd.noColor {
+		colorError = ""
+		colorOk = ""
+		colorNeutral = ""
+		colorMatched = ""
+	}
+
+	if cmd.match != "" && (!cmd.sort || !cmd.unique) {
```
jaroslaw-pieszka, line 102: I wonder if we need so many switches. 
Why not default mode running all necessary checks?
Or switching off particular tests if needed.

cmd/parser/parse.go
```diff
@@ -0,0 +1,356 @@
...
+	cobraCmd.Flags().BoolVarP(&cmd.sort, "priority", "p", false, "Sort rule entries by their priority, in descending priority order.")
+	cobraCmd.Flags().BoolVarP(&cmd.unique, "unique", "u", false, "Display only non duplicated rules. Error entries are not considered for uniqueness.")
+	cobraCmd.Flags().BoolVarP(&cmd.signature, "signature", "s", false, "Mark rules with the mirrored signatures as duplicated. For example aws(PR=*, HR=westeurope) and aws(PR=westeurope, HR=*) are considered duplicated because of having mirrored signatures.")
+	cobraCmd.Flags().BoolVarP(&cmd.noColor, "no-color", "n", false, "Disable use color characters when generating output.")
+	cobraCmd.MarkFlagsOneRequired("entry", "file")
+
+	return cobraCmd
+}
+
+func (cmd *ParseCommand) Run() error {
```
jaroslaw-pieszka, line 92: Have you considered reading from stdin (tool as filter)?
It is easier to call:
`hap parse <ruleset.txt`
than
`hap parse -f ruleset.txt`

cmd/parser/parse_test.go
```diff
@@ -0,0 +1,87 @@
...
+        for _, c := range cases.Case {
+        
+            expected := strings.ReplaceAll(c.ExpectedRule, " ", "")
+            expected = strings.ReplaceAll(expected, "\t", "")
+            expected = strings.ReplaceAll(expected, "\n", "")
+            expected = strings.ReplaceAll(expected, "\r", "")
+            expected = strings.ReplaceAll(expected, "\f", "")
+        
+            entries := ""
+            for i, rule := range c.Rules {
```
jaroslaw-pieszka, line 55: Consider usage of strings.Join()

common/hyperscaler/rules/sorter.go
```diff
@@ -0,0 +1,19 @@
+package rules
+
+import "sort"
+
+func SortRuleEntries(entries []ParsingResult) []ParsingResult {
```
jaroslaw-pieszka, line 5: nit: the number of input attributes could be {0, 1, 2}. We do not restrict number of rules.
From the complexity point of view count sort could be more appropriate.

common/hyperscaler/rules/simple_parser.go
```diff
@@ -0,0 +1,96 @@
...
+)
+
+type SimpleParser struct{
+    
+}
+
+func (g* SimpleParser) Parse(ruleEntry string) (*Rule, error) {
+    outputRule := &Rule{}
+
+    ruleEntry = strings.ReplaceAll(ruleEntry, " ", "")
```
jaroslaw-pieszka, line 15: strings.Map could be handy (and process entry in one run).

cmd/parser/parse.go
```diff
@@ -0,0 +1,356 @@
...
+			}
+
+			negativeSignatureItem, negativeSignatureExists := signatureSet[negativeSignatureKey]
+
+			// if negativeSignatureExists && containsWildcards && cmd.signature {
+			if negativeSignatureExists && cmd.signature {
+				
+				resolvingSignaturePossibleRule := result.Rule.Combine(*negativeSignatureItem.Rule)
+
+			resolvingKey := resolvingSignaturePossibleRule.StringNoLabels()
```
jaroslaw-pieszka, line 215: Something off with indentation.

cmd/parser/parse.go
```diff
@@ -0,0 +1,356 @@
...
+	useGrammar 			 bool
+	ruleFilePath 		 string
+	sort 			 bool
+	unique 			 bool
+	match 			 string
+	signature 			 bool
+	noColor 			 bool
+}
+
+func NewParseCmd() *cobra.Command {
```
jaroslaw-pieszka, line 38: Are we printing all messages to stderr?

common/hyperscaler/rules/rule.go
```diff
@@ -0,0 +1,358 @@
...
+	input         bool
+	output        bool
+	modifiedLabel string
+
+	modifiedLabelName string
+	ApplyLabel        func(r *Rule, labels map[string]string) map[string]string
+}
+
+func (a Attribute) String(r *Rule) any {
+	val := a.getter(r)
```
jaroslaw-pieszka, line 25: Why do we use string representations of bool values instead of bool values?

common/hyperscaler/rules/rule.go
```diff
@@ -0,0 +1,358 @@
...
+var InputAttributes = []Attribute{
+	{
+		Name:        PR_ATTR_NAME,
+		Description: "Platform Region",
+		setter:      setPlatformRegion,
+		getter:      func(r *Rule) string { return r.PlatformRegion },
+		input:       true,
+		output:      false,
+		ApplyLabel: func(r *Rule, labels map[string]string) map[string]string {
+			if labels[HYPERSCALER_LABEL] == "" {
```
jaroslaw-pieszka, line 58: `HYPERSCALER_LABEL` or `PLATFORM_LABEL`

jaroslaw-pieszka, line 58: ApplyLabel is quite similar for both attributes.

common/hyperscaler/rules/rule.go
```diff
@@ -0,0 +1,358 @@
...
+	ASTERISK = "*"
+)
+
+var InputAttributes = []Attribute{
+	{
+		Name:        PR_ATTR_NAME,
+		Description: "Platform Region",
+		setter:      setPlatformRegion,
+		getter:      func(r *Rule) string { return r.PlatformRegion },
+		input:       true,
```
jaroslaw-pieszka, line 55: Do we need these now?
Do we use it, is not it against YAGNI principle?

jaroslaw-pieszka: I have found the usage.
We compute  max allowed number of input attributes.
Still we need to discuss it directly.

PR #1721, opened 2025-02-07
Additional worker node pools E2E test

KsaweryZietara:
**Description**

- extend `update` and `assert` command with additional worker node pools in `skr-tester`.

**Related issue(s)**
See also #952

### DIFF
--- a/testing/e2e/skr-tester/pkg/command/update.go
+++ b/testing/e2e/skr-tester/pkg/command/update.go
@@ -11,16 +11,18 @@ import (
 )
 
 type UpdateCommand struct {
-	cobraCmd             *cobra.Command
-	log                  logger.Logger
-	instanceID           string
-	planID               string
-	updateMachineType    bool
-	updateOIDC           bool
-	updateAdministrators bool
-	customMachineType    string
-	customOIDC           string
-	customAdministrators []string
+	cobraCmd                        *cobra.Command
+	log                             logger.Logger
+	instanceID                      string
+	planID                          string
+	updateMachineType               bool
+	updateOIDC                      bool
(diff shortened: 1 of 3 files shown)


testing/e2e/skr-tester/pkg/command/update.go
```diff
@@ -176,6 +182,48 @@ func (cmd *UpdateCommand) Run() error {
...
+			}
+			fmt.Printf("Update operationID: %s\n", resp["operation"].(string))
+			return nil
+		}
+
+		currentMachineType, err := kcpClient.GetCurrentMachineType(cmd.instanceID)
+		if err != nil {
+			return fmt.Errorf("failed to get current machine type: %v", err)
+		}
+		newAdditionalWorkerNodePools := []map[string]interface{}{
```
jaroslaw-pieszka, line 205: Why do we use hard-wired values? Is it necessary to?

KsaweryZietara: It will simplify integration into our pipelines while also maintaining convention:
- https://github.com/kyma-project/kyma-environment-broker/blob/main/testing/e2e/skr-tester/pkg/command/update.go#L147
- https://github.com/kyma-project/kyma-environment-broker/blob/main/testing/e2e/skr-tester/pkg/command/update.go#L172

However, if you prefer passing it as an argument in the pipelines, I can change that.

jaroslaw-pieszka: Leave it as is, if we need it we will add it.

jaroslaw-pieszka (approved):

PR #1782, opened 2025-02-21
Hap Rules Wiring Part 2

ralikio:
**Description**

- Rules Service as a placeholder for further logic,
- Helm chart test to see if configuration is loaded as expected,
- Test data,
- Yaml configs loading,
- corrected app config file property.

**Related issue(s)**

#1520


common/hyperscaler/rules/app_config_test.go
```diff
@@ -0,0 +1,126 @@
...
+const BROKER_CONTAINER_NAME = "kyma-environment-broker"
+const BROKER_CHART_NAME = "kcp-kyma-environment-broker"
+const VALUES_FILE = HELM_CHART + "/values.yaml"
+const NOTES_FILE = "keb/templates/NOTES.txt"
+
+const ENV_NAME = "APP_HAP_RULE_FILE_PATH"
+const ENV_FILENAME = "hapRule.yaml"
+const ENV_PATH = "/config/" + ENV_FILENAME
+
+// Test for checking if expected format of
```
jaroslaw-pieszka, line 31: Is there something missing in the comment?

common/hyperscaler/rules/app_config_test.go
```diff
@@ -0,0 +1,126 @@
...
+	require.NoError(t, err)
+	require.NotNil(t, resources)
+
+	clientBuilder := fake.NewClientBuilder()
+
+	for filename, res := range resources {
+		if filename == NOTES_FILE {
+			continue
+		}
+		res = strings.Trim(res, "\n")
```
jaroslaw-pieszka, line 67: Could be done in a single call using `strings.Trim`

common/hyperscaler/rules/app_config_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+		// then
+		require.NoError(t, err)
+		require.NotNil(t, deployment)
+
+		envFound := false
+		for _, container := range deployment.Spec.Template.Spec.Containers {
+			if container.Name == BROKER_CONTAINER_NAME {
+				for _, env := range container.Env {
+					if env.Name == ENV_NAME &&
```
jaroslaw-pieszka, line 114: Since there are not so many components and complexity is not an issue I would suggest 
```
envFound := envFound || (env.Name == ENV_NAME && env.Value == ENV_PATH)
```
Without a `break`.

common/hyperscaler/rules/model/config_test.go
```diff
@@ -0,0 +1,43 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestLoad(t *testing.T) {
+	t.Run("should load rules from a file", func(t *testing.T) {
+  
+        // given
+		expectedRules := []string{"rule1", "rule2"}
```
jaroslaw-pieszka, line 14: Could you correct indentation?

ralikio: Lintting issues corrected.

common/hyperscaler/rules/app_config_test.go
```diff
@@ -0,0 +1,118 @@
...
+		}, deployment)
+
+		// then
+		require.NoError(t, err)
+		require.NotNil(t, deployment)
+
+		containers := deployment.Spec.Template.Spec.Containers
+		cIndex := slices.IndexFunc(containers, func(c v1.Container) bool { return c.Name == BROKER_CONTAINER_NAME })
+
+		found := slices.ContainsFunc(containers[cIndex].Env, func(e v1.EnvVar) bool { return e.Name == ENV_NAME && e.Value == ENV_PATH })
```
jaroslaw-pieszka, line 113: So we assume that in deployment there is very specific setting '/config/hapRule.yaml'

common/hyperscaler/rules/model/config_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestLoad(t *testing.T) {
+	t.Run("should load rules from a file", func(t *testing.T) {
```
jaroslaw-pieszka, line 11: In this test case we test libraries (reading file from os and reading yaml).

common/hyperscaler/rules/rules_service.go
```diff
@@ -0,0 +1,27 @@
...
+
+func NewRulesServiceFromFile(rulesFilePath string) (*RulesService, error) {
+	rulesConfig := &model.RulesConfig{}
+
+	if rulesFilePath == "" {
+		return nil, fmt.Errorf("No HAP rules file provided")
+	}
+
+	log.Printf("Parsing rules from file: %s\n", rulesFilePath)
+	rulesConfig, err := rulesConfig.Load(rulesFilePath)
```
jaroslaw-pieszka, line 21: Is this a local value? `rulesConfig`

common/hyperscaler/rules/app_config_test.go
```diff
@@ -0,0 +1,118 @@
...
+	require.NotNil(t, resources)
+
+	clientBuilder := fake.NewClientBuilder()
+
+	for filename, res := range resources {
+		if filename == NOTES_FILE {
+			continue
+		}
+		res = strings.Trim(res, "\n ")
+		if res == "" || res == "\n" || strings.Contains(res, "istio") {
```
jaroslaw-pieszka, line 69: Is this possible to get `res == "\n"` after `strings.Trim(res, "\n ")`?

common/hyperscaler/rules/model/config_test.go
jaroslaw-pieszka: So basically we create slice of strings:
```yaml
rule:
  - rule1
  - rule2
 ```
write it to the file, read it from the file, unmarshal yaml so we get []{"rule1", "rule2"}

File: common/hyperscaler/rules/rules_service.go (hunk 7 above)
Comment (ralikio) [reply common/hyperscaler/rules/rules_service.go:21]: Yes, it will be that way. Subsequent PRs do parsing on the rulesConfig putting the results in RulesService.

File: common/hyperscaler/rules/model/config.go [hunk 9]
```diff
@@ -0,0 +1,26 @@
...
+	"os"
+
+	"gopkg.in/yaml.v2"
+)
+
+type RulesConfig struct {
+	Rules []string `yaml:"rule"`
+}
+
+func (c *RulesConfig) Load(file string) (*RulesConfig, error) {
```
jaroslaw-pieszka, line 14: This is file name of type string. Not file.
But nevertheless I would consider Load using not file name as input, but either interface or []byte collection like in yaml.Unmarshal - to make it easier to test.

common/hyperscaler/rules/rules_service.go
jaroslaw-pieszka: I was just puzzled because this value is lost in current implementation.
It is not returned. It is not passed further.

common/hyperscaler/rules/model/config_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+		// given
+		expectedRules := []string{"rule1", "rule2"}
+		content := "rule:\n"
+
+		for _, rule := range expectedRules {
+			content += "- " + rule + "\n"
+		}
+		tmpfile, err := os.CreateTemp("", "test*.yaml")
```
jaroslaw-pieszka, line 20: Alternatively we could use predefined configs in some `testdata` directory.

common/hyperscaler/rules/model/config.go
ralikio: Changed `file` to `filePath`. I do not want the method to get a byte array because it would just wrap Unmarshal invocation which is excessive.

common/hyperscaler/rules/model/config_test.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestLoad(t *testing.T) {
+	t.Run("should load rules from a file", func(t *testing.T) {
+
+		// given
+		expectedRules := []string{"rule1", "rule2"}
+		content := "rule:\n"
+
```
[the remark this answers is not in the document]
ralikio: Current implementation allows to modify single array of expected rules to modify both test input and assertions.

common/hyperscaler/rules/model/config_test.go
ralikio: I prefer this approach.

common/hyperscaler/rules/model/config.go
jaroslaw-pieszka: And you just wrap two things instead of one.
Os.Read and yaml.Unmarshal.
With such approach you cannot test this little part of logic without having a file.

common/hyperscaler/rules/model/config_test.go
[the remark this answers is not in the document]
jaroslaw-pieszka: And this is the pitfall. 
You modify single source to create input and expected output.

common/hyperscaler/rules/model/config_test.go
jaroslaw-pieszka: This is obvious since you use it.

common/hyperscaler/rules/model/config_test.go
jaroslaw-pieszka: What kind of flaw can we detect?

common/hyperscaler/rules/model/config_test.go
jaroslaw-pieszka: It is easier to understand what is in the file, and how it looks if you can see it in editor/IDE and not infer it from the code that generates config file.

common/hyperscaler/rules/rules_service.go
```diff
@@ -0,0 +1,27 @@
...
+)
+
+type RulesService struct {
+}
+
+func NewRulesServiceFromFile(rulesFilePath string) (*RulesService, error) {
+	rulesConfig := &model.RulesConfig{}
+
+	if rulesFilePath == "" {
+		return nil, fmt.Errorf("No HAP rules file provided")
```
jaroslaw-pieszka, line 17:
```suggestion
		return nil, fmt.Errorf("No HAP rules file name provided")
```

common/hyperscaler/rules/model/config_test.go
piotrmiskiewicz: My 3 cents. I'd love to have the input file in the `given` section, then few asserts in the `then` section, for example:
```
// given
input := "
rule:
  - rule1
  - rule2
"
// save the input to a tmp file

// when
_, err = config.Load(tmpfile.Name())

// then
// asserts 
```
in such example I see what is the input and see two asserts and I can compare using my eyes.
When creating the input file is hidden by some lines of code I do not see it directly. On the other hand, the above solution seems to be simpler (not sure, but maybe shorter)?

common/hyperscaler/rules/model/config_test.go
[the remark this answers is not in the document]
jaroslaw-pieszka: Do we really need this flexibility to extend list of input/output rules in this test?
Will we better cover the `Load` having
```yaml
rule:
  - rule1
  - rule2
  - rule3
 ```

File: common/hyperscaler/rules/model/config.go [hunk 13]
```diff
@@ -0,0 +1,26 @@
...
+
+type RulesConfig struct {
+	Rules []string `yaml:"rule"`
+}
+
+func (c *RulesConfig) Load(filePath string) (*RulesConfig, error) {
+
+	yamlFile, err := os.ReadFile(filePath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read YAML file: %s", err)
```
jaroslaw-pieszka, line 18: I am not sure if this path of execution is tested.

common/hyperscaler/rules/model/config.go
```diff
@@ -0,0 +1,26 @@
...
+	"os"
+
+	"gopkg.in/yaml.v2"
+)
+
+type RulesConfig struct {
+	Rules []string `yaml:"rule"`
+}
+
+func (c *RulesConfig) Load(filePath string) (*RulesConfig, error) {
```
jaroslaw-pieszka, line 14: Why we return the pointer to the receiver of the function?

common/hyperscaler/rules/model/config_test.go
jaroslaw-pieszka: We can detect the changes in error reporting, but only for unmarshaling (os.Read error is not covered).

common/hyperscaler/rules/model/config_test.go
```diff
@@ -0,0 +1,43 @@
...
+		if _, err := tmpfile.Write([]byte(content)); err != nil {
+			t.Fatalf("Failed to write to temp file: %v", err)
+		}
+		if err := tmpfile.Close(); err != nil {
+			t.Fatalf("Failed to close temp file: %v", err)
+		}
+
+		// when
+		var config RulesConfig
+		_, err = config.Load(tmpfile.Name())
```
jaroslaw-pieszka, line 33: And here Load() signature make it harder to get. You skip the return value so at first glimpse we ignore the result of computations apart from error. But we change the state of the config variable.

common/hyperscaler/rules/model/config.go
ralikio: I do not want to mirror libraries' logic by creating a method for every function. It is unnecessary here.

internal/process/provisioning/resolve_creds.go
```diff
@@ -23,12 +24,14 @@ type ResolveCredentialsStep struct {
 	accountProvider  hyperscaler.AccountProvider
 	opStorage        storage.Operations
 	tenant           string
+	rulesService     *rules.RulesService
 }
 
-func NewResolveCredentialsStep(os storage.Operations, accountProvider hyperscaler.AccountProvider) *ResolveCredentialsStep {
+func NewResolveCredentialsStep(os storage.Operations, accountProvider hyperscaler.AccountProvider, rulesService *rules.RulesService) *ResolveCredentialsStep {
```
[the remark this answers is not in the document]
ralikio: I have just used convention that we use in every step for "dependency injection" in our steps. I do not want to add error to the contract of steps' constructors.

jaroslaw-pieszka (dismissed):

jaroslaw-pieszka (approved):

Issue #1734, opened 2025-02-13
High latency on `/runtimes` endpoint - https://github.tools.sap/kyma/backlog/issues/6765


piotrmiskiewicz:
Manifestations:

- high latency for requests sent by KMC
- high latency for `kcp rt`

Probable root cause:
- number of operations stored in database increased significantly:
```
broker=> select count(*) from operations;
 count  
--------
 117093
(1 row)
```
- `ListInstances` and `ListInstancesWithSubaccountStates` use complex query with self join on operations to get lastOperation:
```sql
select o1.data from instances 
join operations o1 on (instances.instance_id = o1.instance_id) 
left join operations o2 on (o2.instance_id = instances.instance_id and o1.created_at < o2.created_at) 
left join subaccount_states ss on (instances.sub_account_id = ss.id) and o2.created_at IS NULL;
```
explain:
```
broker=> explain select o1.data from instances join operations o1 on (instances.instance_id = o1.instance_id) left join operations o2 on (o2.instance_id = instances.instance_id and o1.created_at < o2.created_at) left join subaccount_states ss on (instances.sub_account_id = ss.id) and o2.created_at IS NULL;
                                                          QUERY PLAN                                                           
-------------------------------------------------------------------------------------------------------------------------------
 Hash Right Join  (cost=112005.65..236764.39 rows=1851840 width=990)
   Hash Cond: ((o2.instance_id)::text = (instances.instance_id)::text)
   Join Filter: (o1.created_at < o2.created_at)
   ->  Index Only Scan using operations_by_instance_id_created_at on operations o2  (cost=0.42..30237.03 rows=117945 width=45)
   ->  Hash  (cost=94750.92..94750.92 rows=117945 width=1072)
         ->  Hash Join  (cost=1186.34..94750.92 rows=117945 width=1072)
               Hash Cond: ((o1.instance_id)::text = (instances.instance_id)::text)
               ->  Seq Scan on operations o1  (cost=0.00..93254.45 rows=117945 width=1035)
               ->  Hash  (cost=1155.04..1155.04 rows=2504 width=74)
                     ->  Seq Scan on instances  (cost=0.00..1155.04 rows=2504 width=74)
```
Indexes for operations:
```
broker=> \d operations;
                              Table "public.operations"
         Column          |           Type           | Collation | Nullable | Default 
-------------------------+--------------------------+-----------+----------+---------
 id                      | character varying(255)   |           | not null | 
[...]

PR #1784, opened 2025-02-24
Reduce logging while introducing new queries based on `last_operation_id`

jaroslaw-pieszka:
**Description**

- Filter not logged
- added keyword to make filtering easier

**Related issue(s)**

#1734

### DIFF
--- a/internal/storage/postsql/read.go
+++ b/internal/storage/postsql/read.go
@@ -705,7 +705,7 @@ func (r readSession) GetActiveInstanceStats() ([]dbmodel.InstanceByGlobalAccount
 		States: []dbmodel.InstanceState{dbmodel.InstanceNotDeprovisioned},
 	}
 
-	slog.Info("Calling - GetActiveInstanceStats", "filter", filter)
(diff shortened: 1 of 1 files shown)


internal/storage/postsql/read.go
```diff
@@ -705,7 +705,7 @@ func (r readSession) GetActiveInstanceStats() ([]dbmodel.InstanceByGlobalAccount
 		States: []dbmodel.InstanceState{dbmodel.InstanceNotDeprovisioned},
 	}
 
-	slog.Info("Calling - GetActiveInstanceStats", "filter", filter)
+	slog.Info("[last_operation_id] Calling GetActiveInstanceStats()")
```
ralikio, line 708: Can you use logger flags/fields?

MarekMichali (approved):

internal/storage/postsql/read.go
jaroslaw-pieszka: These logs will be removed tomorrow, the day after tomorrow probably.

Issue #423, opened 2024-02-01
Extend Kyma Provisioning Parameters to allow multiple OIDC definitions


kwiatekus:
**Description**

Adjust set of [input parameters](https://github.com/kyma-project/kyma-environment-broker/blob/4ec8fc184786376a95b365d394dbd3b65b9a14ab/resources/keb/files/swagger.yaml#L2087)  of Kyma Service Instance Provisioning so that user can provide multiple OIDC configs (design `oidc` paramater so that in accepts a single config (for backwards compatibility) or an array).

Extend the [OIDC](https://github.com/kyma-project/kyma-environment-broker/blob/4ec8fc184786376a95b365d394dbd3b65b9a14ab/resources/keb/files/swagger.yaml#L2056) schema so that user could also define [requiedClaims](https://github.com/gardener/gardener-extension-shoot-oidc-service/blob/master/docs/usage/openidconnects.md?plain=1#L81C2-L81C86) (key-value pairs) that are essential for secure GH workflow access.

**AC**
Phase 1: migration - Size/M
 - [x] coordinate migration with SRE, see below
 - [x] Implement solutions for migration  (see: https://github.com/kyma-project/kyma/issues/18305#issuecomment-2612019987)

Phase 2: feature implementation - Size/L
 - [x] User can define multiple OIDC configs
 - [x] User can define required claims per OIDC config
 - [x] API proposal
   - [x] must be backward compatible
 - [x] Additional OIDC proposal Implementation (see: https://github.com/kyma-project/kyma/issues/18305#issuecomment-2128866460)
 - [x] Kubeconfig "context" enhancement to show all OIDCs that user sets. Default set to array[0]
https://github.com/kyma-project/kyma-environment-broker/pull/1936
    - [x] INFO: Busola will make use of that to allow user to select on of them.  Busola ticket: https://github.com/kyma-project/busola/issues/3653
 - [x] E2E test 
 - [x] Test load_current_config feature with OIDC data with object vs array. Check is existing values are available to user in the BTP form.
 - [x] Deploy: (only possible if Phase 1 fininshed)
   - [x] KIM ticket closed and deployed to all envs: https://github.com/kyma-project/infrastructure-manager/issues/803
   - [x] dev + ERS update
   - [x] stage + ERS update
   - [x] prod + ERS update 

 
**Reasons**

It is required to configure access to freshly created clusters via additional "workflow" OIDC
https://github.com/kyma-project/kyma/issues/18305

[...]

PR #1804, opened 2025-02-27
Populate additionalOIDC field during provisioning and update

MarekMichali:
**Description**

Part of the phase 1: migration.  We duplicate the main OIDC to additionalOIDC[0] when the feature flag is enabled.

- Copy OIDC value to additionalOIDC when enabled
- Add a feature flag that enables/disables writing to the mainOIDC
- Add a feature flag that enables/disables writing to the additionalOIDC
- Add tests that use different feature flags enabled

**Related issue(s)**

See #423

### DIFF
--- a/internal/process/provisioning/create_runtime_resource_step.go
+++ b/internal/process/provisioning/create_runtime_resource_step.go
@@ -298,13 +298,20 @@ func (s *CreateRuntimeResourceStep) createKubernetesConfiguration(operation inte
 		}
 	}
 
-	return imv1.Kubernetes{
-		Version: ptr.String(s.config.KubernetesVersion),
-		KubeAPIServer: imv1.APIServer{
-			OidcConfig:           oidc,
-			AdditionalOidcConfig: &[]gardener.OIDCConfig{oidc},
-		},
+	kubernetesConfig := imv1.Kubernetes{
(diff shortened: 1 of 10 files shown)


internal/process/provisioning/create_runtime_resource_step.go
```diff
@@ -298,13 +298,21 @@ func (s *CreateRuntimeResourceStep) createKubernetesConfiguration(operation inte
...
-		KubeAPIServer: imv1.APIServer{
-			OidcConfig:           oidc,
-			AdditionalOidcConfig: nil,
-		},
+	kubernetesConfig := imv1.Kubernetes{
+		Version:       ptr.String(s.config.KubernetesVersion),
+		KubeAPIServer: imv1.APIServer{},
+	}
+
+	if s.config.UseMainOIDC {
```
jaroslaw-pieszka, line 306: So we have two separate settings, hence four combinations are possible?

MarekMichali: Yes, but the scenario where both flags are false shouldn't happen. It is in our responsibility to not set both to false. I can add a check for this if you think we need this.

jaroslaw-pieszka (approved):

PR #1812, opened 2025-03-03
Timeouts in EDP and deprovisioning steps using `RetryOperations` helper function

jaroslaw-pieszka:
**Description**

- EDP registration step
- EDP deregistration step 
- Deprovisioning initialization step

**Related issue(s)**

### DIFF
--- a/internal/process/deprovisioning/edp_deregistration.go
+++ b/internal/process/deprovisioning/edp_deregistration.go
@@ -43,6 +43,11 @@ func NewEDPDeregistrationStep(os storage.Operations, is storage.Instances, clien
 	return step
 }
 
+const (
+	edpRetryInterval = 10 * time.Second
+	edpRetryTimeout  = 30 * time.Minute
+)
+
 func (s *EDPDeregistrationStep) Name() string {
 	return "EDP_Deregistration"
 }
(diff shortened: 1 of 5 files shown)


internal/process/provisioning/edp_registration.go
```diff
@@ -89,30 +94,29 @@ func (s *EDPRegistrationStep) Run(operation internal.Operation, log *slog.Logger
...
 	return newOp, 0, nil
 }
 
 func (s *EDPRegistrationStep) handleError(operation internal.Operation, err error, log *slog.Logger, msg string) (internal.Operation, time.Duration, error) {
 	log.Warn(fmt.Sprintf("%s: %s", msg, err))
-
-	if kebError.IsTemporaryError(err) {
-		since := time.Since(operation.UpdatedAt)
-		if since < time.Minute*30 {
+	if s.config.Required {
```
ralikio, line 106: Can we check temporary error first to avoid duplication in lines 107 and 113 as it is done in the original code?

jaroslaw-pieszka: We can, but I decided that this is easier to understand despite duplication.
We have clearly separated paths for different configs.
Either way the duplication remains. 
This needs to be addressed in timeouts refactoring story.

internal/process/deprovisioning/edp_deregistration.go
```diff
@@ -98,16 +103,13 @@ func (s *EDPDeregistrationStep) Run(operation internal.Operation, log *slog.Logg
 
 func (s *EDPDeregistrationStep) handleError(operation internal.Operation, err error, log *slog.Logger, msg string) (internal.Operation, time.Duration, error) {
 	if kebError.IsTemporaryError(err) {
-		since := time.Since(operation.UpdatedAt)
-		if since < time.Minute*30 {
-			log.Warn(fmt.Sprintf("request to EDP failed: %s. Retry...", err))
-			return operation, 10 * time.Second, nil
-		}
+		return s.operationManager.RetryOperationWithoutFail(operation, s.Name(), "request to EDP failed", edpRetryInterval, edpRetryTimeout, log, err)
```
ralikio, line 106: As I understand it if we returned `return operation, 10 * time.Second, nil` in previous implementation in that if statement and the step would timeout using operation timeout mechanism then eventually the operation would be finished in failed state (see staged_manager.go - Execute method)? If my understanding is correct then we should use `RetryOperation` here.

internal/process/provisioning/edp_registration.go
jaroslaw-pieszka:
```go
	if kebError.IsTemporaryError(err) {
                if s.config.Required {
                   RetryOperationWithoutFail()
                } else {
                   RetryOperation()
		}
	}

	if s.config.Required {
        	return s.operationManager.OperationFailed(operation, msg, err, log)
	} else {
		log.Warn(fmt.Sprintf("Step %s failed. Step is not required. Skip step.", s.Name()))
		return operation, 0, nil
        }
```

internal/process/deprovisioning/edp_deregistration.go
piotrmiskiewicz: as I remember, there was an agreement that we should only fail if k8s cluster still exists, in other cases - use "executedNotCompleted" logic

jaroslaw-pieszka: Good point. So in this step we should not fail.

PR #1845, opened 2025-03-13
Change interface of the method Match

piotrmiskiewicz:
**Description**

- more friendly interface of Match method
- remove asterisks

**Related issue(s)**

### DIFF
--- a/common/hyperscaler/rules/rule.go
+++ b/common/hyperscaler/rules/rule.go
@@ -232,7 +232,7 @@ func (r *Rule) SignatureWithValues() string {
 }
 
 func (r *Rule) MirroredSignature() string {
-	return r.SignatureWithSymbols(ATTRIBUTE_WITH_VALUE, "")
+	return r.SignatureWithSymbols(ATTRIBUTE_WITH_VALUE, "*")
 }
 
 // SignatureWithSymbols returns the signature of the rule with the given symbols with a format similar to the input:
--- a/cmd/parser/rules/test-cases.yaml
+++ b/cmd/parser/rules/test-cases.yaml
@@ -22,29 +22,17 @@ cases:
   - preview
   - free
   expected: |
-    ->     OK  aws                                                # hyperscalerType:
-    aws                            
-    ->     OK  aws(PR=cf-eu11)-> EU                               #
-    euAccess: true, hyperscalerType: aws    
-    ->     OK  azure                                              #
-    hyperscalerType: azure                          
-    ->     OK  azure(PR=cf-ch20)->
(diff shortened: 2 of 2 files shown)


common/hyperscaler/rules/rule.go
```diff
@@ -236,13 +236,17 @@ func (r *Rule) Combine(rule Rule) *Rule {
 }
 
 func (r *Rule) SignatureWithValues() string {
-	signature := r.Plan
+	signature := r.Plan + L_PAREN
 
-	for _, attr := range InputAttributes {
-		signature += attr.Name + SIGNATURE_ATTR_SEPARATOR
+	for i, attr := range InputAttributes {
```
jaroslaw-pieszka, line 241: I am not sure if we need a loop here - there are two InputAttributes.

common/hyperscaler/rules/rule.go
```diff
@@ -236,13 +236,17 @@ func (r *Rule) Combine(rule Rule) *Rule {
...
+		signature += attr.Name + EQUAL
 		checkValue := attr.Getter(r)
 		signature += getAttrValueSymbol(checkValue, ASTERISK, checkValue)
+		if i < len(InputAttributes)-1 {
```
jaroslaw-pieszka, line 245: In the previous line we can get ASTERISK, so this should adjusted as well.

common/hyperscaler/rules/rules_service.go
```diff
@@ -92,6 +92,20 @@ func (rs *RulesService) parse(rulesConfig *RulesConfig) *ParsingResults {
 	return results
 }
 
+// Match finds the matching rule for the given provisioning attributes and provide the set of labels, which must be used to find proper secret binding.
+func (rs *RulesService) MatchProvisioningAttributes(provisioningAttributes *ProvisioningAttributes) (Result, bool) {
+	var result Result
+	found := false
+	for _, parsingResult := range rs.Parsed.Results {
+		if !parsingResult.HasParsingErrors() && parsingResult.Rule.Matched(provisioningAttributes) {
```
jaroslaw-pieszka, line 100: Does it mean that we check parsing errors each time we try to calculate labels?
If there are parsing errors we shouldn't try to match. 
In case of KEB we should fail to start.
In case of CLI command we should fail as well.
If KEB is running we should assume there is no parsing errors. Unless we are very, very defensive.

To sum up.
In KEB we parse ruleset once. If it is not ok, we abort.
If it is ok, we can match provisioning attributes for each instance creation assuming that rules are valid, unambiguous, unique.

jaroslaw-pieszka:
Done in scope of others tickets. See #1520

PR #1855, opened 2025-03-15
HAP verified by cmd/broker tests

piotrmiskiewicz:
**Description**

- cmd/broker tests now uses gardener fake client with a list of SecretBindings with proper labels

**Related issue(s)**


cmd/broker/testdata/old-sap-converged-cloud-region-mappings.yaml
```diff
@@ -1,2 +1,3 @@
 cf-eu20-staging:
   - eu-de-1
+  - eu-de-2
```
piotrmiskiewicz, line 3: this file contains a mapping between platform regions and available openstack regions. It is taken into account in the catalog endpoint which is used by the Pivotal framework. So if the unit test tries to create sapconverged-cloud runtime, but the Platform region does not match the cf-eu20-staging - it fails

cmd/broker/suite_test.go
```diff
@@ -339,25 +342,57 @@ func fixConfig() *Config {
...
+		"sb-aws": {
+			"hyperscalerType": "aws",
+		},
+		"sb-gcp": {
+			"hyperscalerType": "gcp",
+		},
+		"sb-gcp_cf-sa30": {
+			"hyperscalerType": "gcp_cf-sa30",
+		},
+		"sb-aws-shared": {
```
jaroslaw-pieszka, line 359: Do we need azure shared?

jaroslaw-pieszka (approved):

PR #1856, opened 2025-03-17
Adjust sec file to not scan tests

MarekMichali:
**Description**

The test shouldn't be scanned. For more details why this prefix is needed see https://github.tools.sap/kyma/test-infra/issues/577#issuecomment-10953720.

\

**Related issue(s)**


szwedm (approved):
/lgtm

PR #1859, opened 2025-03-17
KimConfig not used any more

jaroslaw-pieszka:
**Description**

Since the migration to KIM is done, we need to clean up the code needed during migration

- KimConfig structure and its usage removed.

**Related issue(s)**


piotrmiskiewicz:
what about chart definition?

jaroslaw-pieszka:
> what about chart definition?

I was going to add this in the next PR, but I will it add it immediately to this one.

piotrmiskiewicz (approved):

PR #1865, opened 2025-03-18
Handle object and list type OIDC input param, add requiredClaims support

MarekMichali:
**Description**

This PR should be merged after SRE finishes the migration to additionalOIDC. 

This PR adds the ability for KEB to handle either OIDC object or OIDC list from input params during provisioning or updating and adds support for requiredClaims and groupsPrefix. Check the [comment](https://github.com/kyma-project/kyma-environment-broker/issues/423#issuecomment-2753473297) for supported scenarios (and verify that no scenario is missing). 

The feature is not on a feature flag, because as long as the new schema is not exposed, from the users POV nothing will change. He won't be able to send a OIDC list, he will be blocked on schema validation. The requiredClaims feature is on the same feature flag as exposing the new schema but we will already validate it if provided in input params. The support for multiple OIDCs in kubeconfig will be added in a separate PR.

- Remove OIDC migration related feature flags
- Handle object and list type OIDC input param
- Handle requiredClaims from OIDC input
- Handle groupsPrefix from OIDC input
- Add tests
- RequiredClaims on expose new schema feature flag
- GroupsPrefix  on expose new schema feature flag

**Related issue(s)**

See #423


common/runtime/model.go
```diff
@@ -158,63 +158,178 @@ type OIDCConfigDTO struct {
 	SigningAlgs    []string `json:"signingAlgs" yaml:"signingAlgs"`
 	UsernameClaim  string   `json:"usernameClaim" yaml:"usernameClaim"`
 	UsernamePrefix string   `json:"usernamePrefix" yaml:"usernamePrefix"`
+	RequiredClaims []string `json:"requiredClaims" yaml:"requiredClaims"`
 }
 
 const oidcValidSigningAlgs = "RS256,RS384,RS512,ES256,ES384,ES512,PS256,PS384,PS512"
 
-func (o *OIDCConfigDTO) IsProvided() bool {
+func (o *OIDCsDTO) IsProvided() bool {
```
jaroslaw-pieszka, line 166: Easy to misspell. But I do not get any proposal.

internal/process/provisioning/create_runtime_resource_step_test.go
```diff
@@ -235,7 +271,82 @@ func TestCreateRuntimeResourceStep_OIDC_MixedCustom(t *testing.T) {
...
+	runtime := imv1.Runtime{}
+	err = cli.Get(context.Background(), client.ObjectKey{
+		Namespace: "kyma-system",
+		Name:      operation.RuntimeID,
+	}, &runtime)
+	assert.NoError(t, err)
+	assert.Equal(t, expectedOIDCConfig, (*runtime.Spec.Shoot.Kubernetes.KubeAPIServer.AdditionalOidcConfig)[0])
+}
+
+func TestCreateRuntimeResourceStep_HandleEmptyOIDCList(t *testing.T) {
```
MarekMichali, line 291: For PK

common/runtime/model.go
```diff
@@ -158,63 +158,178 @@ type OIDCConfigDTO struct {
...
 				break
 			}
 		}
 	}
+}
 
-	if len(errs) > 0 {
-		err := fmt.Errorf(strings.Join(errs, ", "))
-		return err
+func (o *OIDCsDTO) validateSigningAlgsForIndex(signingAlgs []string, index int, errs *[]string) {
```
jaroslaw-pieszka, line 286: Please consider not passing index.
Just returning error and decorating this error with index on the caller level.

common/runtime/model.go
```diff
@@ -158,63 +158,178 @@ type OIDCConfigDTO struct {
...
+}
+
+func (o *OIDCsDTO) validateRequiredClaims(requiredClaims []string, errs *[]string) {
+	if len(requiredClaims) != 0 {
+		for _, claim := range requiredClaims {
+			if !strings.Contains(claim, "=") {
+				*errs = append(*errs, fmt.Sprintf("requiredClaims must be in claim=value format, invalid claim: %s", claim))
+				continue
+			}
+			parts := strings.SplitN(claim, "=", 2)
```
jaroslaw-pieszka, line 305: We do not detect more than one `=`.

MarekMichali: > func strings.SplitN(s string, sep string, n int) []string
> SplitN slices s into substrings separated by sep and returns a slice of the substrings between those separators.
> 
> The count determines the number of substrings to return:
> 
> n > 0: at most n substrings; the last substring will be the unsplit remainder;

`key=value=junk` will be split to `key`, `value=junk`. The user may provide some hash value that contains `=` so thats why we do the split only once on the first found `=`

jaroslaw-pieszka: My bad. It is not an issue. I did not get it properly.

common/runtime/model.go
jaroslaw-pieszka: To avoid so many upper case letters: `OidcDTO` or this previous name `OIDCConfigDTO` or `OIDConnectDTO`

MarekMichali: I think that `OIDConnectDTO` is the way. `OIDCConfigDTO` is already in use and thats why a new name was needed.

cmd/broker/provisioning.go
```diff
@@ -106,7 +106,7 @@ func NewProvisioningProcessingQueue(ctx context.Context, provisionManager *proce
 		// postcondition: operation.KymaResourceName, operation.RuntimeResourceName is set
 		{
 			stage:     createRuntimeStageName,
-			step:      provisioning.NewCreateRuntimeResourceStep(db.Operations(), db.Instances(), k8sClient, cfg.InfrastructureManager, defaultOIDC),
+			step:      provisioning.NewCreateRuntimeResourceStep(db.Operations(), db.Instances(), k8sClient, cfg.InfrastructureManager, defaultOIDC, cfg.Broker.UseAdditionalOIDCSchema),
```
jaroslaw-pieszka, line 109: Could we add this configuration setting to InfrastructureManager instead of Broker?
Since this affects Runtime CR.

common/runtime/model.go
```diff
@@ -156,67 +156,173 @@ func (p AutoScalerParameters) Validate(planMin, planMax int) error {
...
+func (o *OIDCConnectDTO) IsProvided() bool {
 	if o == nil {
 		return false
 	}
-	if o.ClientID == "" && o.IssuerURL == "" && o.GroupsClaim == "" && o.UsernamePrefix == "" && o.UsernameClaim == "" && len(o.SigningAlgs) == 0 {
-		return false
+	if o.OIDCConfigDTO != nil && (o.OIDCConfigDTO.ClientID != "" || o.OIDCConfigDTO.IssuerURL != "" || o.OIDCConfigDTO.GroupsClaim != "" || o.OIDCConfigDTO.UsernamePrefix != "" || o.OIDCConfigDTO.UsernameClaim != "" || len(o.OIDCConfigDTO.SigningAlgs) > 0 || len(o.OIDCConfigDTO.RequiredClaims) > 0 || o.OIDCConfigDTO.GroupsPrefix != "") {
+		return true
+	}
+	if o.List != nil {
```
jaroslaw-pieszka, line 176: Why not to cut it short:
```go
return o.List != nil
```

MarekMichali: Because I didn't see it. I will change it.

internal/broker/instance_update.go
```diff
@@ -240,6 +240,14 @@ func (b *UpdateEndpoint) processUpdateParameters(instance *internal.Instance, de
 			logger.Error(fmt.Sprintf("unable to unmarshal parameters: %s", err.Error()))
 			return domain.UpdateServiceSpec{}, fmt.Errorf("unable to unmarshal parameters")
 		}
+		if !b.config.UseAdditionalOIDCSchema && params.OIDC != nil {
```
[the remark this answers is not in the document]
MarekMichali: This code is temporary as this is a feature flag but I can do something about it.

jaroslaw-pieszka (approved):

jaroslaw-pieszka (approved):

PR #1872, opened 2025-03-19
Subscription Service - post-parse processing redesigned

jaroslaw-pieszka:
**Description**

- Uniqueness check redone
- Ambiguity check redone
- Matching redone
- Unit tests added

**Related issue(s)**


common/hyperscaler/rules/rules_service.go
```diff
@@ -64,13 +68,63 @@ func NewRulesServiceFromSlice(rules []string, enabledPlans *broker.EnablePlans)
...
 	}
 
-	rs.ParsedRuleset = rs.parse(rulesConfig)
+	rs.ParsedRuleset = rs.process(rulesConfig)
+	_, rs.ValidRules, rs.ValidationInfo = rs.processAndValidate(rulesConfig)
 	return rs, nil
 }
 
-func (rs *RulesService) parse(rulesConfig *RulesConfig) *ParsingResults {
+func (rs *RulesService) processAndValidate(rulesConfig *RulesConfig) (bool, *ValidRuleset, *ValidationErrors) {
```
piotrmiskiewicz, line 76: do we need the bool returned?

common/hyperscaler/rules/ruleset.go
```diff
@@ -0,0 +1,143 @@
...
+
+func (vr *ValidRule) Match(provisioningAttributes *ProvisioningAttributes) bool {
+	if !vr.Plan.Match(provisioningAttributes.Plan) {
+		return false
+	}
+
+	return vr.matchRegions(provisioningAttributes)
+}
+
+func (vr *ValidRule) matchRegions(provisioningAttributes *ProvisioningAttributes) bool {
```
piotrmiskiewicz, line 42: in general it is about matching input parameters, I know we have only regions as input params, but maybe the name could be about input params, nit regions

common/hyperscaler/rules/ruleset.go
```diff
@@ -0,0 +1,143 @@
...
+func (r *ValidRule) toResult(provisioningAttributes *ProvisioningAttributes) Result {
+	hyperscalerType := provisioningAttributes.Hyperscaler
+	if r.PlatformRegionSuffix {
+		hyperscalerType += "_" + provisioningAttributes.PlatformRegion
+	}
+	if r.HyperscalerRegionSuffix {
+		hyperscalerType += "_" + provisioningAttributes.HyperscalerRegion
+	}
+
+	return Result{
```
piotrmiskiewicz, line 62: one note, which is not directly about the PR but can be easy implemented here. Let's add new field to the Result, which contains the rule as a string. So the step (ResolveSubscriptionSecret) calls MatchProvisioningAttributesWithValidRuleset gets not only 3 values used to create label selector. It gets also the original rule just for logging it. At the end of the day, the operator can look at KEB logs and find exaclty which rule was taken into account.

common/hyperscaler/rules/rules_service.go
jaroslaw-pieszka: We do not. It is redundant however more explicit.
Removed in the next push.

common/hyperscaler/rules/ruleset.go
jaroslaw-pieszka: Pushed.

piotrmiskiewicz (approved):

Issue #1820, opened 2025-03-06
SKR-tester improvements


MarekMichali:
**AC**
- [x] wait after kcp kubeconfig so there is time to grant access to the SKR for the technical user (prevent this https://some-github/some-org/some-repo/actions/runs/11766021/job/51145999) https://github.com/kyma-project/kyma-environment-broker/pull/1883
  - [x] when we can't connect with the kubeconfig from kcp cli, it should be logged 
- [x] kcp rt status must be printed for every operation when has end state or timeout etc (this was reverted by Marek from nodejs version, we need to bring it back) https://github.com/kyma-project/kyma-environment-broker/pull/1885
- [x] run kcp rt --events for the instance when a test fails https://github.com/kyma-project/kyma-environment-broker/pull/1886
- [x] build-runtime-* workflow, allow to run every plan separately [#51](https://github.tools.sap/kyma/gophers-pipelines/pull/51), [#53](https://github.tools.sap/kyma/gophers-pipelines/pull/53), [#54](https://github.tools.sap/kyma/gophers-pipelines/pull/54)
- [x] add missing skr-tester-gcp-integration workflow [#52](https://github.tools.sap/kyma/gophers-pipelines/pull/52)
- [ ] ~~allow operations to fail on KEB timeout, not on test timeout. TEst timeout == 110% of the KEB timeout.
https://github.com/kyma-project/kyma-environment-broker/pull/1898~~
- [x] wrap all errors returned by kcp cli https://github.com/kyma-project/kyma-environment-broker/pull/1903
- [x] use the latest kcp-cli release (See issues 49 in our test repo) [#56](https://github.tools.sap/kyma/gophers-pipelines/pull/56)

MarekMichali:
More details about issue AC 4: https://github.com/orgs/community/discussions/8774

PR #1898, opened 2025-03-27
Increase operation timeout in SKR-tester

KsaweryZietara:
**Description**

- increase `check_operation` timeout in SKR-tester to allow operations to fail on KEB timeout, not on test timeout.

**Related issue(s)**
See also #1820

### DIFF
--- a/testing/e2e/skr-tester/pkg/command/check_operation.go
+++ b/testing/e2e/skr-tester/pkg/command/check_operation.go
@@ -35,7 +35,7 @@ func NewCheckOperationCommand() *cobra.Command {
 
 	cobraCmd.Flags().StringVarP(&cmd.instanceID, "instanceID", "i", "", "Instance ID of the specific instance.")
 	cobraCmd.Flags().StringVarP(&cmd.operationID, "operationID", "o", "", "OperationID of the specific operation.")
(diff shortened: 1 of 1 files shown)


testing/e2e/skr-tester/pkg/command/check_operation.go
```diff
@@ -35,7 +35,7 @@ func NewCheckOperationCommand() *cobra.Command {
 
 	cobraCmd.Flags().StringVarP(&cmd.instanceID, "instanceID", "i", "", "Instance ID of the specific instance.")
 	cobraCmd.Flags().StringVarP(&cmd.operationID, "operationID", "o", "", "OperationID of the specific operation.")
-	cobraCmd.Flags().DurationVarP(&cmd.timeout, "timeout", "t", 40*time.Minute, "Timeout for the operation to finish.")
+	cobraCmd.Flags().DurationVarP(&cmd.timeout, "timeout", "t", 25*time.Hour, "Timeout for the operation to finish.")
```
MarekMichali, line 38: Is this value equal to 110% of our operation timeout?

KsaweryZietara: Operation timeout is equal to [24h](https://github.com/kyma-project/kyma-environment-broker/blob/main/cmd/broker/main.go#L90), 110% * 24h = 26.4h, but I thought 25 is enough. Should i change it to 26.4h?

PR #1940, opened 2025-04-07
Avoid unnecessary error on response label creation

piotrmiskiewicz:
**Description**

- when runtime ID is empty  - the kubeconfig is not yet created so we can't return kubeconfig URL and it is not an error
- do not log error when RuntimeID is empty

**Related issue(s)**


szwedm (approved):
/lgtm

PR #1973, opened 2025-04-14
Add ID types to logs

KsaweryZietara:
**Description**

- add ID types to logs where the type is missing.

### DIFF
--- a/internal/broker/bind_create.go
+++ b/internal/broker/bind_create.go
@@ -153,7 +153,7 @@ func (b *BindEndpoint) bind(ctx context.Context, instanceID, bindingID string, d
 
 	lastOperation, err := b.operationsStorage.GetLastOperation(instance.InstanceID)
 	if err != nil {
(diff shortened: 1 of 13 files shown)


internal/broker/bind_create.go
```diff
@@ -153,7 +153,7 @@ func (b *BindEndpoint) bind(ctx context.Context, instanceID, bindingID string, d
 
 	lastOperation, err := b.operationsStorage.GetLastOperation(instance.InstanceID)
 	if err != nil {
-		return domain.Binding{}, apiresponses.NewFailureResponse(fmt.Errorf("failed to get last operation for instance %s", instanceID), http.StatusInternalServerError, fmt.Sprintf("failed to get last operation %s", instanceID))
+		return domain.Binding{}, apiresponses.NewFailureResponse(fmt.Errorf("failed to get last operation for instance %s", instanceID), http.StatusInternalServerError, fmt.Sprintf("failed to get last operation for instance %s", instanceID))
```
MarekMichali, line 156: I think it would be better to use the word `instanceID` like in the other places.

MarekMichali: For other teams that use our logs it may be instance!=instanceID, they could assume that instance=runtimeID.

MarekMichali (approved):

Issue #1924, opened 2025-04-02
Enable g6 and g4 machine types for aws plan in the two regions from subset of all zones


PK85:
**Description**

Working on adding G4 machine types for aws: [ticket](https://github.com/kyma-project/kyma-environment-broker/issues/1854#issuecomment-2771879005), we figured out that two region has more than 3 zones we support:
- 5 zones: "us-east-1": "abcdf" | g6 is not suported in f
- 4 zones: "us-west-2": "abcd" | g4 not supported in d

We created ticket to infrastructure manager to allow KEB to use any 3 ouf of all zones for given worker pools: [KIM ticket](https://github.com/kyma-project/infrastructure-manager/issues/804)

**AC**
- [x] 1) [KIM ticket](https://github.com/kyma-project/infrastructure-manager/issues/804) is closed and deployed to all environments
- [x] 2) Extend out current region <-> machineType mapping to include zones too, see details with proposal
https://github.com/kyma-project/kyma-environment-broker/pull/1998 
https://github.com/kyma-project/kyma-environment-broker/pull/2008
https://github.com/kyma-project/kyma-environment-broker/pull/2026
- [x] 3) Support g6 in "us-east-1" and support g4 in "us-west-2" 
https://github.tools.sap/kyma/management-plane-config/pull/6470
  - [x] 3.1) Regions where zones are not provided in the mapping, uses current mechanism
  - [x] 3.2) dev [#6449](https://github.tools.sap/kyma/management-plane-config/pull/6449), [#6498](https://github.tools.sap/kyma/management-plane-config/pull/6498)
  - [x] 3.3) stage [#6455](https://github.tools.sap/kyma/management-plane-config/pull/6455), [#6501](https://github.tools.sap/kyma/management-plane-config/pull/6501)
  - [x] 3.4) prod [#6472](https://github.tools.sap/kyma/management-plane-config/pull/6472), [#6515](https://github.tools.sap/kyma/management-plane-config/pull/6515)
- [x] 4) Support regions with only 1 or 2 zones. User can use given machine type in such a region only if HA is set to FALSE. If user marked HA true then fail with message like: "This region has no 3 zones. If you want to use this machine type you need to set HA to false".
- [x] 5) when HA is false, KEB always randomly selects one zone only. Regardless if region has 2,3 or 4 zones. Must select one from the configured ones. E.g.       - us-eastwest13 has only a & b. That means it should select one, a or b.
[...]

PR #1998, opened 2025-04-24
Extend regions supporting machine with zone mapping

KsaweryZietara:
**Description**

- introduce `zoneMapping` feature flag,
- support the new `regionsSupportingMachine` configuration format when the corresponding feature flag is enabled,
- use zone mapping during the creation and update of the RuntimeCR,
- validate if region have at least 3 zones if HA is enabled.

**Related issue(s)**
See also #1924


piotrmiskiewicz:
I have an idea, to not pass zoneMapping over lot of files/structs/methods:
1. Create a "class" which is responsible for creating additional workers with one method: CreateAdditionalWorkers. The list of params will be shorter - all constant values will be a part of that class, the list of args would be (values, currentAdditionalWorkers, additionalWorkerNodePools, zones, planID). The rest of args can be moved to the constructor. Then, the class is instantiated in the main.go and passed to provisioning/update steps.
2. The CreateRuntimeResourceStep is not aware of zonemapping boolean flag
3. The same for UpdateRuntimeStep

piotrmiskiewicz (approved):

PR #2003, opened 2025-04-30
Disable machine type update

KsaweryZietara:
**Description**

- introduce `disableMachineTypeUpdate` feature flag,
- disable machine type update in additional worker node pools,
- change machine type description.

**Related issue(s)**
See also #1924


common/runtime/model.go
```diff
@@ -534,3 +534,14 @@ func (a AdditionalWorkerNodePool) ValidateHAZonesUnchanged(currentAdditionalWork
 	}
 	return nil
 }
+
+func (a AdditionalWorkerNodePool) ValidateMachineTypesUnchanged(currentAdditionalWorkerNodePools []AdditionalWorkerNodePool) error {
```
jaroslaw-pieszka, line 538: I wonder if it would be feasible to separate duties here.

1. Predicate checking if any machine type has been changed (possibly with positive logic)
2. Action - returning error

jaroslaw-pieszka (approved):

PR #2022, opened 2025-05-09
Regions configurable

piotrmiskiewicz:
**Description**

- regions are configurable per plan/platformRegion
- if there is no given platformRegion defined and no "default"entry - plan is not available
- redesign code which generates schemas

Testing:
 - call catalog endpoint in the main branch KEB (or latest release), save to the file
 - call catalog endpoint in the PR, save to the file
 - compare files

**Related issue(s)**


piotrmiskiewicz:
generating catalog response could be done in the provisioning_test.go (first test)

internal/broker/schemas.go
```diff
@@ -0,0 +1,347 @@
...
+		providerSpec:                providerSpec,
+		defaultOIDCConfig:           defaultOIDCConfig,
+		cfg:                         cfg,
+		ingressFilteringFeatureFlag: ingressFilteringEnabled,
+		ingressFilteringPlans:       ingressFilteringPlans,
+	}, nil
+}
+
+func (s *SchemaService) Plans(plans PlansConfig, platformRegion string, cp pkg.CloudProvider) map[string]domain.ServicePlan {
+	awsCreateSchema := s.AWSSchema(platformRegion, false)
```
jaroslaw-pieszka, line 43: This repetitive pattern is error prone. It is easy to use the wrong boolean flag in the invocation.
I do not have perfect solution but using single function per plan returning two schemas (create and update) and encapsulating calls with boolean `update` appeals to me. 
But this is very subjective.

internal/broker/schemas.go
```diff
@@ -0,0 +1,347 @@
...
+		OwnClusterPlanID:        defaultServicePlan(OwnClusterPlanID, OwnClusterPlanName, plans, ownClusterCreateSchema, ownClusterUpdateSchema),
+		PreviewPlanID:           defaultServicePlan(PreviewPlanID, PreviewPlanName, plans, previewCreateSchema, previewUpdateSchema),
+		BuildRuntimeAWSPlanID:   defaultServicePlan(BuildRuntimeAWSPlanID, BuildRuntimeAWSPlanName, plans, buildRuntimeAWSCreateSchema, buildRuntimeAWSUpdateSchema),
+		BuildRuntimeGCPPlanID:   defaultServicePlan(BuildRuntimeGCPPlanID, BuildRuntimeGCPPlanName, plans, buildRuntimeGCPCreateSchema, buildRuntimeGCPUpdateSchema),
+		BuildRuntimeAzurePlanID: defaultServicePlan(BuildRuntimeAzurePlanID, BuildRuntimeAzurePlanName, plans, buildRuntimeAzureCreateSchema, buildRuntimeAzureUpdateSchema),
+		SapConvergedCloudPlanID: defaultServicePlan(SapConvergedCloudPlanID, SapConvergedCloudPlanName, plans, sapConvergedCloudCreateSchema, sapConvergedCloudUpdateSchema),
+	}
+	// remove plans if there is no any region defined for given platform region (for example strict list of platform regions for sap converged cloud)
+	for id := range outputPlans {
```
jaroslaw-pieszka, line 83: Nit picking - wouldn't be easier to add these plans with regions instead of adding all and deleting some?
Subjective again, we need to discuss it briefly (30s).

internal/broker/schemas.go
```diff
@@ -0,0 +1,347 @@
...
+		// if trial or Own cluster - do not check regions
+		// freemium - always enabled
+		if id == TrialPlanID || id == OwnClusterPlanID || id == FreemiumPlanID {
```
jaroslaw-pieszka, line 86: Could we extract this logic as separate predicate function?

internal/provider/configuration/provider.go
```diff
@@ -0,0 +1,88 @@
...
+func NewProviderSpec(r io.Reader) (*ProviderSpec, error) {
+	data := &dto{}
+	d := yaml.NewDecoder(r)
+	err := d.Decode(data)
+	return &ProviderSpec{
+		data: *data,
+	}, err
+}
+
+func (p *ProviderSpec) RegionDisplayName(cp runtime.CloudProvider, region string) string {
```
jaroslaw-pieszka, line 36: Following our direct discussion about missing display names - we could handle it here (returning region as "default").

internal/provider/configuration/provider.go
```diff
@@ -0,0 +1,88 @@
...
+	dto := p.findRegion(cp, region)
+	if dto == nil {
+		return region
```
jaroslaw-pieszka, line 39: We default to region silently. Should we report in logs such case?

internal/provider/configuration/provider.go
```diff
@@ -0,0 +1,88 @@
...
+	return nil
+}
+
+func (p *ProviderSpec) Validate(provider runtime.CloudProvider, region string) error {
+	if dto := p.findRegion(provider, region); dto != nil {
+		if dto.Zones == nil || len(dto.Zones) == 0 {
+			return fmt.Errorf("region %s for provider %s has no zones defined", region, provider)
+		}
+		if dto.DisplayName == "" {
+			return fmt.Errorf("region %s for provider %s has no display name defined", region, provider)
```
jaroslaw-pieszka, line 83: In `RegionDisplayName` we fallback to region as display name. 
I would consider at most warning and fallback, possibly just fallback but not an error.

cmd/broker/update.go
```diff
@@ -16,13 +18,17 @@ import (
...
-	cfg Config, kcpClient client.Client, logs *slog.Logger, workersProvider *workers.Provider) *process.Queue {
+	cfg Config, kcpClient client.Client, logs *slog.Logger, workersProvider *workers.Provider, schemaService *broker.SchemaService) *process.Queue {
 
 	trialRegionsMapping, err := provider.ReadPlatformRegionMappingFromFile(cfg.TrialRegionMappingFilePath)
 	if err != nil {
 		fatalOnError(err, logs)
 	}
 
+	regions, err := provider.ReadPlatformRegionMappingFromFile(cfg.TrialRegionMappingFilePath)
+	valuesProvider := provider.NewPlanSpecificValuesProvider(cfg.InfrastructureManager.MultiZoneCluster, cfg.InfrastructureManager.DefaultTrialProvider,
```
jaroslaw-pieszka, line 29: May be we should just pass cfg.InfrastructerManager as a whole reducing number of parameters passed.

internal/provider/aws.go
```diff
@@ -61,11 +67,14 @@ var awsZones = map[string]string{
 
 func (p *AWSInputProvider) Provide() internal.ProviderValues {
 	zonesCount := p.zonesCount()
-	zones := p.zones()
 	region := DefaultAWSRegion
 	if p.ProvisioningParameters.Parameters.Region != nil {
 		region = *p.ProvisioningParameters.Parameters.Region
 	}
+	zones := AWSZones(region, p.ZonesProvider.RandomZones(pkg.AWS, region, zonesCount))
```
jaroslaw-pieszka, line 74: `zonesCount` is computed based on hard-wired `DefaultAWSMultiZoneCount` - should it be so?

internal/provider/configuration/provider.go
```diff
@@ -0,0 +1,101 @@
...
+	if dto == nil {
+		return []string{}
+	}
+	return dto.Zones
+}
+
+func (p *ProviderSpec) RandomZones(cp runtime.CloudProvider, region string, zonesCount int) []string {
+	availableZones := p.Zones(cp, region)
+	rand.Shuffle(len(availableZones), func(i, j int) { availableZones[i], availableZones[j] = availableZones[j], availableZones[i] })
+	if zonesCount > len(availableZones) {
```
jaroslaw-pieszka, line 69: It is quite obvious what happens when `zonesCount <= len(availableZones)`.
We pick randomly `zonesCount` zones out of `availableZones`.

But when  `zonesCount > len(availableZones)' we just return all available zones (randomly ordered).

Is this scenario possible? 
Should we allow it?

piotrmiskiewicz: Sap Converged Cloud sometimes have 2 zones, the code for zones in the main branch is: 
```go
func ZonesForSapConvergedCloud(region string, zonesCount int) []string {
	zones, found := sapConvergedCloudZones[region]
	if !found {
		zones = "a"
		zonesCount = 1
	}

	availableZones := strings.Split(zones, "")
	rand.Shuffle(len(availableZones), func(i, j int) { availableZones[i], availableZones[j] = availableZones[j], availableZones[i] })
	if zonesCount > len(availableZones) {
		// get maximum number of zones for region
		zonesCount = len(availableZones)
	}

	availableZones = availableZones[:zonesCount]

	var generatedZones []string
	for _, zone := range availableZones {
		generatedZones = append(generatedZones, fmt.Sprintf("%s%s", region, zone))
	}
	return generatedZones
}
```

jaroslaw-pieszka (approved):

PR #2030, opened 2025-05-14
Support multiple OIDCs for KEB

berkanercan:
**Description**

- Introduced a new system to populate Authorization-policy CRs with multiple oidc tenants, if they are defined in values.yaml.
- For stage we are going to support two tenants. Thats why we need this change. Here is the related issue: https://github.tools.sap/kyma/backlog/issues/3899
- ...

**Related issue(s)**

### DIFF
--- a/resources/keb/templates/authorization-policy.yaml
+++ b/resources/keb/templates/authorization-policy.yaml
@@ -60,7 +60,14 @@ spec:
     from:
       - source:
           requestPrincipals:
+          {{- $oidcs := .Values.oidcs | default list }}
+          {{- if gt (len $oidcs) 0 }}
+          {{- range $i, $p := $oidcs }}
+          - {{ $p.issuer }}/*
+          {{- end }}
+          {{- else }}
           - {{ tpl .Values.oidc.issuer $ }}/*
+          {{- end }}
     when:
     - key: request.auth.claims[scp]
       values:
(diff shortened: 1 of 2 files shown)


resources/keb/templates/authorization-policy.yaml
```diff
@@ -60,7 +60,14 @@ spec:
     from:
       - source:
           requestPrincipals:
+          {{- $oidcs := .Values.oidcs | default list }}
+          {{- if gt (len $oidcs) 0 }}
+          {{- range $i, $p := $oidcs }}
+          - {{ $p.issuer }}/*
+          {{- end }}
```
ebensom, line 67: I would call it as `.Values.oidc.issuers`, already having a default values of empty list in the values.yaml (which has boolean value of false):

```suggestion
          {{- if .Values.oidc.issuers }}
          {{- range $i, $p := .Values.oidc.issuers }}
          - {{ $p}}/*
          {{- end }}
```

resources/keb/values.yaml
```diff
@@ -330,6 +330,8 @@ oidc:
     orchestrations: orchestrationsAdmin
     viewer: runtimeViewer
 
+oidcs: []
```
ebensom, line 333: instead, I would just have a new issuers list with empty default under the existing oidc config:

```suggestion
oidc:
  issuers: []
```

PR #2031, opened 2025-05-14
Describe Additional Worker Node Pool Autoscaler

szwedm:
**Description**

Describe behavior of additional worker node pool autoscaler when the **autoScalerMin** value is set to 0.

**Related issue(s)**

See #2019

### DIFF
--- a/docs/user/04-40-additional-worker-node-pools.md
+++ b/docs/user/04-40-additional-worker-node-pools.md
@@ -53,7 +53,8 @@ With high availability enabled, resources are distributed across three zones to 
 In this scenario, you must set **autoScalerMin** to at least `3`.
 
 If high availability is disabled, all resources are placed in a single, randomly selected zone. In this case, you can set **autoScalerMin** to 0 and **autoScalerMax** to `1`, which helps reduce costs. 
-However, it is not recommended for production environments. 
+However, it is not recommended for production environments. Setting the **autoScalerMin** value to 0 results in an automatic removal of the additional worker nodes depending on the current workload in the cluster. 
+If there are no user's workloads deployed onto the nodes associated with the given additional worker node pool, the nodes should be removed automatically in around 30 mins.
 
 See the following JSON example without the `additionalWorkerNodePools` list:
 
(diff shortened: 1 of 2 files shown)


docs/user/04-40-additional-worker-node-pools.md
```diff
@@ -53,7 +53,8 @@ With high availability enabled, resources are distributed across three zones to
 In this scenario, you must set **autoScalerMin** to at least `3`.
 
 If high availability is disabled, all resources are placed in a single, randomly selected zone. In this case, you can set **autoScalerMin** to 0 and **autoScalerMax** to `1`, which helps reduce costs. 
-However, it is not recommended for production environments. 
+However, it is not recommended for production environments. Setting the **autoScalerMin** value to 0 results in an automatic removal of the additional worker nodes depending on the current workload in the cluster. 
+If there are no user's workloads deployed onto the nodes associated with the given additional worker node pool, the nodes should be removed automatically in around 30 mins.
```
IwonaLanger, line 57:
```suggestion
If there are no user workloads deployed onto the nodes associated with the given additional worker node pool, the nodes are removed automatically in around 30 minutes.
```

IwonaLanger: `Should` doesn't sound definitive. Is there an option that the nodes aren't removed in the described circumstances?

szwedm: The autoscaler behavior is highly dependent on the current cluster state and workloads. We observed the scaling down of nodes during tests in our clusters, but our test environment may differ greatly from the clients environments. That's the reason we cannot guarantee nodes scale down in ~30 mins. We can add a reference to Gardener documentation about autoscaling: https://gardener.cloud/docs/gardener/autoscaling/shoot_autoscaling/

IwonaLanger: So if we're only uncertain of the duration time, why don't we phrase it like this:

If there are no user workloads deployed onto the nodes associated with the given additional worker node pool, the nodes are removed automatically. The process should take around 30 minutes.

IwonaLanger (approved):

PR #2093, opened 2025-06-05
Redesign of creating additional worker pools - zones

piotrmiskiewicz:
**Description**

- redesign creation of additional worker pools to not create full zone names - use common function instead

**Related issue(s)**


cmd/broker/provisioning_test.go
```diff
@@ -1218,7 +1220,7 @@ func TestProvisioning_ClusterParameters(t *testing.T) {
 		},
 		"Production Multi-AZ AWS": {
 			planID:                       broker.AWSPlanID,
-			region:                       "us-east-1",
+			region:                       "sa-east-1",
```
piotrmiskiewicz, line 1223: this region has 3 zones

KsaweryZietara (approved):

Issue #1997, opened 2025-04-23
Monitor what parameters are sent to KEB. Gather ones that do not fit our create and update schema.


PK85:
**Description**

After collecting data, we can decide to deny calls where payloads contain data we do not work with. Those requests could be rejected already on CIS side, not reaching KEB by using  "additonalPropoerties false" on root for create and update schemas.

**AC**
- test if "additonalPropoerties false" on root for create and update schema works as expected (We expect CIS will reject such call. Try to test it with mocking data in the webbrowser)
- Monitor what parameters are sent to KEB. Gather ones that do not fit our create and update schema.

MarekMichali:
https://json-schema.org/understanding-json-schema/reference/object#additionalproperties

KsaweryZietara:
After setting `additionalProperties` to false in the mocked schema and adding a test field to the JSON, BTP Cockpit displays the following error message:
```
No additional properties allowed but property test is set
```

<img width="731" alt="Image" src="https://github.com/user-attachments/assets/79dbf582-ec07-4d9b-923e-6cc4eca49655" />

KsaweryZietara:
| Pages | Requests per page | Total download time     | Data size | KEB memory |
| :-------: |:-------------------:|:-------------------------:|------------------|------------|
| 150 | 10,000            | 2 minutes 40 seconds    | File with 1.5 mln request (~1 GB)     |  Increased from 0.08 GiB to 0.12 GiB       |
| 15 | 100,000           | 1 minute 40 seconds     | File with 1.5 mln request (~1 GB)     |      Increased from 0.08 GiB to 0.72 GiB     |

KsaweryZietara:
Monitoring for parameters sent to KEB that do not fit to our create/update schema is now enabled in the following environments:
- [X] dev
- [X] stage
- [ ] prod

The issue is now moved to "Blocked" until sufficient data has been collected.

KsaweryZietara:
Moved to Blocked. Next Actions:
- Schedule meeting with Cloud Connector team.
- Organize a meeting to discuss the breaking change.

PK85:
Status:

Meeting Sep 8 2025 with cloud orchestrator team. We aligned that issue is on cloud orch side and they will fix it by implementing this: https://github.com/SAP/crossplane-provider-btp/issues/126

PR #2099, opened 2025-06-10
Reject unsupported parameters

KsaweryZietara:
**Description**

- introduce `rejectUnsupportedParameters` feature flag,
- if `rejectUnsupportedParameters` is `true`, set `additionalProperties` to `false`.

**Related issue(s)**
See also #1997


jaroslaw-pieszka (approved):

resources/keb/values.yaml
```diff
@@ -167,6 +167,8 @@ broker:
   operationTimeout: "7h"
   # Port for the broker HTTP server
   port: "8080"
+  # If true, rejects requests that contains parameters that are not defined in schemas
```
grego952, line 170:
```suggestion
  # If true, rejects requests that contain parameters that are not defined in schemas
```

PR #2146, opened 2025-06-25
Quota limit check

KsaweryZietara:
**Description**

- introduce `checkQuotaLimit` feature flag,
- fetch numer of assigned quota from CIS `provisioning-service`,
- validates during provisioning that the assigned quota for the subaccount is not exceeded.

**Related issue(s)**


internal/quota/client.go
```diff
@@ -0,0 +1,92 @@
...
+	var response Response
+	if err := json.Unmarshal(bodyBytes, &response); err != nil {
+		return 0, fmt.Errorf("while unmarshaling response: %w", err)
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return 0, fmt.Errorf("API error: %s", response.Error.Message)
+	}
+
+	if response.Plan != planName {
```
jaroslaw-pieszka, line 87: This is kind of superdefensive. We explicitly call with `planName` parameter, so why do we check the response for planName?

PR #2152, opened 2025-06-27
Quota whitelist

KsaweryZietara:
**Description**

- 

**Related issue(s)**


resources/keb/values.yaml
```diff
@@ -326,6 +326,11 @@ profiler:
 
 providersConfiguration: {}
 
+# List of subaccount IDs that have unlimited quota for Kyma runtimes.
```
jaroslaw-pieszka, line 329: Should we mention here, that this should be used only on DEV environment?

jaroslaw-pieszka (approved):

resources/keb/values.yaml
KsaweryZietara: Some of the tests are also executed on STAGE environment.

PR #2161, opened 2025-07-02
KCP cli shows actions

KsaweryZietara:
**Description**

- move the `Action` struct to `common/runtime/model.go`,
- add a new `action` query parameter,
- update the runtime endpoint to return actions when the `action` parameter is set to `true`.

See [#7316](https://github.tools.sap/kyma/backlog/issues/7316#issuecomment-13085098)

### DIFF
--- a/internal/runtime/handler.go
+++ b/internal/runtime/handler.go
@@ -258,7 +258,7 @@ func (h *Handler) getRuntimes(w http.ResponseWriter, req *http.Request) {
 		if actions {
 			actions, err := h.actionsDb.ListActionsByInstanceID(dto.InstanceID)
 			if err != nil {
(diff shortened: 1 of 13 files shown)


internal/runtime/handler.go
```diff
@@ -252,6 +255,15 @@ func (h *Handler) getRuntimes(w http.ResponseWriter, req *http.Request) {
 				return
 			}
 		}
+		if actions {
+			actions, err := h.actionsDb.ListActionsByInstanceID(dto.InstanceID)
+			if err != nil {
+				h.logger.Warn(fmt.Sprintf("unable to apply actions: %s", err.Error()))
```
jaroslaw-pieszka, line 261:
```suggestion
				h.logger.Warn(fmt.Sprintf("unable to list actions: %s", err.Error()))
```

PR #2192, opened 2025-07-14
Fix panic when updating an empty oidc list

MarekMichali:
**Description**

When the user tries to update an empty oidc list (list with 0 element) with and oidc object, which all fields are empty, the validation handles it as "oidc not provided" and tries just to update the first element of the additionalOidc list in runtimeCR. But that element doesn't exists as this list is empty and KEB panics.

This PR expand the input parameters validation, where any oidc value is handled as "oidc provided". This mitigates the scenarios as described above.

- More strict validation
- Change oidc validation HTTP error code from 422 to 400

**Related issue(s)**

### DIFF
--- a/internal/broker/instance_update.go
+++ b/internal/broker/instance_update.go
@@ -368,8 +368,13 @@ func (b *UpdateEndpoint) processUpdateParameters(ctx context.Context, instance *
 	}
 
 	if params.OIDC.IsProvided() {
-		instance.Parameters.Parameters.OIDC = params.OIDC
-		updateStorage = append(updateStorage, "OIDC")
+		if params.OIDC.OIDCConfigDTO.ClientID != "" || params.OIDC.OIDCConfigDTO.IssuerURL != "" || params.OIDC.OIDCConfigDTO.GroupsClaim != "" ||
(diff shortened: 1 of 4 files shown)


internal/broker/instance_update.go
```diff
@@ -368,8 +368,13 @@ func (b *UpdateEndpoint) processUpdateParameters(ctx context.Context, instance *
 	}
 
 	if params.OIDC.IsProvided() {
-		instance.Parameters.Parameters.OIDC = params.OIDC
-		updateStorage = append(updateStorage, "OIDC")
+		if params.OIDC.List != nil || (params.OIDC.OIDCConfigDTO.ClientID != "" || params.OIDC.OIDCConfigDTO.IssuerURL != "" || params.OIDC.OIDCConfigDTO.GroupsClaim != "" ||
```
jaroslaw-pieszka, line 371: How about extracting 
```go
o.OIDCConfigDTO.ClientID == "" && o.OIDCConfigDTO.IssuerURL == "" && o.OIDCConfigDTO.GroupsClaim == "" &&
		o.OIDCConfigDTO.UsernamePrefix == "" && o.OIDCConfigDTO.UsernameClaim == "" && len(o.OIDCConfigDTO.SigningAlgs) == 0 &&
		len(o.OIDCConfigDTO.RequiredClaims) == 0 && o.OIDCConfigDTO.GroupsPrefix == "" && o.OIDCConfigDTO.EncodedJwksArray == ""
```
and using negation of this here?

PR #2208, opened 2025-07-17
Switch from Provisioning to Entitlements Service

KsaweryZietara:
**Description**

- fetch numer of assigned quota from CIS `entitlements-service` instead of `provisioning-service`,
- adjust configuration to `entitlement-service`,
- update docs.

**Related issue(s)**

### DIFF
--- a/internal/quota/client_test.go
+++ b/internal/quota/client_test.go
@@ -14,43 +14,51 @@ import (
 )
 
 func TestGetQuota_Success(t *testing.T) {
-	// given
-	response := Response{
-		Plan:  "aws",
-		Quota: 2,
+	tests := []struct {
+		name     string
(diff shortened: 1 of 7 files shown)


internal/quota/client_test.go
```diff
@@ -14,43 +14,51 @@ import (
...
-	// given
-	response := Response{
-		Plan:  "aws",
-		Quota: 2,
+	tests := []struct {
+		name     string
+		response map[string]any
+		expected int
+	}{
+		{
```
jaroslaw-pieszka, line 22: Since we get float and use int to round the value, please add test case with real va

jaroslaw-pieszka (approved):

mmitoraj (approved):

PR #2234, opened 2025-07-29
Not used logging removed to reduce noise

jaroslaw-pieszka:
**Description**

- Logging for queue workers removed where it does not add any value

**Related issue(s)**

### DIFF
--- a/internal/process/queue_test.go
+++ b/internal/process/queue_test.go
@@ -53,8 +53,6 @@ func TestWorkerLogging(t *testing.T) {
 		// then
 		stringLogs := cw.buf.String()
 		t.Log(stringLogs)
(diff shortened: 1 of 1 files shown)


internal/process/queue.go
```diff
@@ -61,12 +61,10 @@ func (q *Queue) ShutDown() {
 }
 
 func (q *Queue) Run(stop <-chan struct{}, workersAmount int) {
-	q.log.Info(fmt.Sprintf("starting %d worker(s), queue length is %d", workersAmount, q.queue.Len()))
 	for i := 0; i < workersAmount; i++ {
 		q.waitGroup.Add(1)
 
 		workerLogger := q.log.With("workerId", i)
-		workerLogger.Info(fmt.Sprintf("starting worker with id %d", i))
```
MarekMichali, line 68: I think workerLogger definition could be also removed.

MarekMichali (approved):

PR #2268, opened 2025-08-14
Plan updates - docs update

piotrmiskiewicz:
**Description**

- describe a plan updates with a details about list of upgradableToPlans field

**Related issue(s)**


docs/contributor/03-80-plan-updates.md
```diff
@@ -18,14 +18,17 @@ plansConfiguration:
       - build-runtime-gcp
 ```
 
+> [!NOTE]
+> The `upgradableToPlans` field is a list of plan names that the current plan can be upgraded to. If the value is an empty list (or not defined) or the list contains only the name of the configured plan (`gcp` in the above example), the plan cannot be updated and the `plan_updateable` in the response of `catalog` endpoint is set to false.
```
IwonaLanger, line 22:
```suggestion
> The **upgradableToPlans** field is a list of plan names to which you can upgrade the current plan. If the value is an empty (or not defined) list, or the list contains only the name of the configured plan (like `gcp` in the above example), the plan cannot be updated, and the **plan_updateable** field in the response of the `catalog` endpoint is set to `false`.
```

PR #2349, opened 2025-09-17
Validate number of discovered zones

KsaweryZietara:
**Description**

- during provisioning and update, perform validation by retrieving a random AWS secret from Gardener and using it to fetch the number of available zones per machine type.

**Related issue(s)**
See also [#7745](https://github.tools.sap/kyma/backlog/issues/7745)

### DIFF
--- a/internal/hyperscalers/aws/client.go
+++ b/internal/hyperscalers/aws/client.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
+	"time"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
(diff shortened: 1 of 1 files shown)


internal/broker/instance_create.go
```diff
@@ -359,6 +373,31 @@ func (b *ProvisionEndpoint) validate(ctx context.Context, details domain.Provisi
 		)
 	}
 
+	var awsClient aws.Client
+	if b.providerSpec.ZonesDiscovery(pkg.CloudProviderFromString(values.ProviderType)) {
```
piotrmiskiewicz, line 377: if AZure/GCP discovery zones are enabled (for example by accident) we have no implementation for it - all provisioning fails, let's discuss

PR #2444, opened 2025-10-23
Machines availability endpoint

KsaweryZietara:
**Description**

- create endpoint which returns machines availability.

**Related issue(s)**
See also [#7746](https://github.tools.sap/kyma/backlog/issues/7746)

### DIFF
--- a/internal/machinesavailability/handler.go
+++ b/internal/machinesavailability/handler.go
@@ -17,6 +17,11 @@ import (
 	"github.com/kyma-project/kyma-environment-broker/internal/subscriptions"
 )
 
+const (
+	machinesAvailabilityPath  = "/oauth/machines_availability"
+	highAvailabilityThreshold = 3
+)
+
 type ProvidersData struct {
 	Providers []Provider `json:"providers"`
 }
@@ -61,7 +66,7 @@ func NewHandler(
 }
 
 func (h *Handler) AttachRoutes(router *httputil.Router) {
-	router.HandleFunc("/oauth/machines_availability", h.getMachinesAvailability)
+	router.HandleFunc(machinesAvailabilityPath, h.getMachinesAvailability)
 }
 
 func (h *Handler) getMachinesAvailability(w http.ResponseWriter, req *http.Request) {
@@ -91,6 +96,7 @@ func (h *Handler) getMachinesAvailability(w http.ResponseWriter, req *http.Reque
 		for _, machineType := range machineTypes {
 			var family string
 			if provider == runtime.AWS {
+				// For AWS, machine types follow the pattern "<family>.<size>".
 				parts := strings.SplitN(machineType, ".", 2)
 				family = parts[0]
 			} else {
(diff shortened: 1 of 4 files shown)


internal/machinesavailability/handler.go
```diff
@@ -0,0 +1,183 @@
...
+
+		machineTypes := h.providerSpec.MachineTypes(provider)
+		machineFamilies := make(map[string]string)
+		for _, machineType := range machineTypes {
+			var family string
+			if provider == runtime.AWS {
+				parts := strings.SplitN(machineType, ".", 2)
+				family = parts[0]
+			} else {
+				httputil.WriteErrorResponse(w, http.StatusInternalServerError, fmt.Errorf("%s not supported", provider))
```
jaroslaw-pieszka:
```suggestion
				httputil.WriteErrorResponse(w, http.StatusInternalServerError, fmt.Errorf("%s provider not supported", provider))
```

internal/machinesavailability/handler.go
```diff
@@ -0,0 +1,183 @@
...
+					return
+				}
+
+				count, err := client.AvailableZonesCount(context.Background(), machineType)
+				if err != nil {
+					httputil.WriteErrorResponse(w, http.StatusInternalServerError, err)
+					return
+				}
+
+				highAvailability := count >= 3
```
jaroslaw-pieszka: Could we extract this magic value as constant?

internal/machinesavailability/handler.go
```diff
@@ -0,0 +1,183 @@
...
+					Name:             region,
+					HighAvailability: highAvailability,
+				})
+			}
+
+			providerEntry.MachineTypes = append(providerEntry.MachineTypes, machineTypeEntry)
+		}
+
+		sort.Slice(providerEntry.MachineTypes, func(i, j int) bool {
+			return providerEntry.MachineTypes[i].Name < providerEntry.MachineTypes[j].Name
```
jaroslaw-pieszka: Is it relevant that we take case into account (when sorting)?

internal/machinesavailability/handler.go
```diff
@@ -0,0 +1,183 @@
...
+		})
+
+		providersData.Providers = append(providersData.Providers, providerEntry)
+	}
+
+	httputil.WriteResponse(w, http.StatusOK, providersData)
+}
+
+func (h *Handler) clientCredentials(provider string) (string, string, error) {
```
jaroslaw-pieszka: I wonder if dividing this function into three separate part would make sense:
- getting rule
- getting secret binding
- extracting credentials (this is as a matter of fact already done).
Rationale:
 - making it easier to migrate from secret bindings to credential bindings.

internal/machinesavailability/handler.go
```diff
@@ -0,0 +1,183 @@
...
+				Regions: []Region{},
+			}
+
+			regions := regionSupportingMachine.SupportedRegions(machineType)
+			if len(regions) == 0 {
+				regions = h.providerSpec.Regions(provider)
+			}
+
+			for _, region := range regions {
+				client, err := h.clientFactory.New(context.Background(), accessKeyID, secretAccessKey, region)
```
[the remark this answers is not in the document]
KsaweryZietara: Yes, an AWS library client requires a region to be specified when it is created. See [NewClient](https://github.com/kyma-project/kyma-environment-broker/blob/main/internal/hyperscalers/aws/client.go#L49-L55), [newAWSConfig](https://github.com/kyma-project/kyma-environment-broker/blob/main/internal/hyperscalers/aws/client.go#L130C6-L136).
The creation takes approximately 0.5 ms, so it’s not performance-intensive.

Issue #2557, opened 2025-11-26
Introduce new AES-GCM mode for write operations


jaroslaw-pieszka:
As first stage we need to introduce FIPS compliant AES mode for new write operations, supporting obsolete mode for read operations.

AC:
- [x] we can recognise which data is encrypted using CFB and which using GCM mode, we decided to introduce additional column for each involved DB table
- [x] all write operations store data encrypted with GCM mode **if proper feature flag is enabled**
- [x] read operations are mode-aware and decrypt data accordingly
- [x] we present statistics in the logs (during start-up) showing how many records/rows are encrypted with obsolete CFB mode

PR #2558, opened 2025-11-26
Write data using AES-GCM

jaroslaw-pieszka:
**Description**

- Encrypt using GCM
- Decrypt using both CFB and GCM
- persist which mode was used
- log statistics
- use feature flags to control 

**Related issue(s)**

#2557


jaroslaw-pieszka:
In the scope of another PR.

PR #2672, opened 2025-12-23
Fix step retries after operation requeuing

### LINKED ISSUE #2569: Operation fails on first retry attempt
**Description**

An operation is failed instead of retried.

**Expected result**

Retry operation accordingly to the passed retry interval and retry timeout arguments.
Check and list in the comment, how many steps has retries in the provisioning operations (check update and delete operations as well)

**Actual result**

Failed operation on a first retry attempt without the retry interval and retry timeout consideration.

**Troubleshooting**

The root cause of the issue is already existing retry record in the operation manager's retryTimestamps map:
https://github.com/kyma-project/kyma-environment-broker/blob/0c8974a7e0d2301b0264c4edce36c78be98ef082/internal/process/operation_manager.go#L22
https://github.com/kyma-project/kyma-environment-broker/blob/0c8974a7e0d2301b0264c4edce36c78be98ef082/internal/process/steps/discover_available_zones.go#L38
https://github.com/kyma-project/kyma-environment-broker/blob/0c8974a7e0d2301b0264c4edce36c78be98ef082/internal/process/operation_manager.go#L101

If there were step retries in a previous operation processing run, the timeout for the step is calculated from the previous retries, so it is already reached.

szwedm:
**Description**

Fixes step retries when the operation is added back to the queue. 

**Related issue(s)**
Fixes #2569

### DIFF
--- a/internal/process/operation_manager_test.go
+++ b/internal/process/operation_manager_test.go
@@ -7,7 +7,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/kyma-project/kyma-environment-broker/internal/fixture"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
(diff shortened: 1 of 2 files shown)


piotrmiskiewicz:
closeing because of better solution discussed with the team, see: https://github.com/kyma-project/kyma-environment-broker/issues/2683

Issue #2780, opened 2026-01-27
gocyclo - bind_create


jaroslaw-pieszka:
https://wiki.one.int.sap/wiki/spaces/kyma/pages/5825828909/Linter+analyisis

internal/broker/bind_create.go:97:1: cyclomatic complexity 34 of func `(*BindEndpoint).bind` is high (> 30) (gocyclo)
func (b *BindEndpoint) bind(ctx context.Context, instanceID, bindingID string, details domain.BindDetails, asyncAllowed bool) (domain.Binding, error) {

Investigate and refactor.

PR #2784, opened 2026-01-28
Cyclo complexity in `bind_create.go`

jaroslaw-pieszka:
**Description**

- Reduced to 20

**Related issue(s)**

#2780


internal/broker/bind_create.go
```diff
@@ -165,20 +165,41 @@ func (b *BindEndpoint) bind(ctx context.Context, instanceID, bindingID string, d
 		return domain.Binding{}, apiresponses.NewFailureResponse(errors.New(message), http.StatusBadRequest, message) // Agreed with Provisioning API team to return 400
 	}
 
+	binding, apiResponse, found := b.searchDbForBinding(err, instanceID, bindingID, expirationSeconds)
```
szwedm: What do you think about adding a method like `isBindingEmpty` instead of `found` variable?

Issue #2802, opened 2026-02-02
staticcheck linter


jaroslaw-pieszka:
https://wiki.one.int.sap/wiki/spaces/kyma/pages/5825828909/Linter+analyisis
minor issues but a lot of them
399 issues:
* staticcheck: 399

Here you can find settings for linter:
- https://golangci-lint.run/docs/linters/configuration/#staticcheck
probably copilot can do that and there will be less after finishing other linters.
- here you can find settings for golangci-yaml to e.g. exclude some files/ directories etc: https://golangci-lint.run/docs/configuration/file/
- So the configuration is 2 levels 
  - specific for linter
  - global configuration e.g. exclusions (those can be used globally or apply only with selected linter)

jaroslaw-pieszka:
In future use:
```yaml
    - staticcheck
  settings:
    staticcheck:
      checks: ["all", "-QF1008"]  # Disable specific staticcheck rules
```

PR #2830, opened 2026-02-05
Staticcheck fixes

jaroslaw-pieszka:
**Description**

- Generated by linter

**Related issue(s)**

#2802

### DIFF
--- a/internal/storage/storage.go
+++ b/internal/storage/storage.go
@@ -70,7 +70,7 @@ func NewMemoryStorage() BrokerStorage {
 		operation:         op,
 		subaccountStates:  ss,
 		instance:          memory.NewInstance(op, ss),
-		events:            events.New(events.Config{}, NewInMemoryEvents()),
+		events:            events.New(events.Config{}, newInMemoryEvents()),
 		instancesArchived: memory.NewInstanceArchivedInMemoryStorage(),
 		bindings:          memory.NewBinding(),
 		actions:           memory.NewAction(),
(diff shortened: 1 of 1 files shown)


cmd/broker/suite_test.go
```diff
@@ -211,7 +211,7 @@ seedRegions:
 	}
 
 	for _, version := range additionalKymaVersions {
-		kebCfg.ObjectMeta.Labels[fmt.Sprintf("runtime-version-%s", version)] = "true"
+		kebCfg.Labels[fmt.Sprintf("runtime-version-%s", version)] = "true"
 	}
```
MarekMichali: MHO, flattening structs in the whole PR removes context for the given field. Also, there is a risk that at some point we could need to unflatten it if a new field appears at the same level where the flattening was done.

internal/storage/storage.go
```diff
@@ -87,8 +87,7 @@ func NewInMemoryEvents() *inMemoryEvents {
 	}
 }
 
-func (_ *inMemoryEvents) RunGarbageCollection(pollingPeriod, retention time.Duration) {
-	return
+func (*inMemoryEvents) RunGarbageCollection(pollingPeriod, retention time.Duration) {
 }
```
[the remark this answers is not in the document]
jaroslaw-pieszka: I had similar impression at some point.

cmd/broker/suite_test.go
jaroslaw-pieszka: Done in the next push.

PR #2839, opened 2026-02-09
Cleaning up context after each iteration in infinite loop

jaroslaw-pieszka:
**Description**

- Implicit cance() invocation

**Related issue(s)**


internal/kymacustomresource/kyma_cr_updater.go
```diff
@@ -82,6 +82,7 @@ func (u *Updater) Run() error {
 				retryRequired = true
 			}
 		}
+		cancel()
```
MarekMichali: What if we jump earlier into an if statement and `continue` the loop? We won't call this cancel and we will create a new context.

Issue #2898, opened 2026-02-20
gocyclo - instance_update


jaroslaw-pieszka:
https://wiki.one.int.sap/wiki/spaces/kyma/pages/5825828909/Linter+analyisis
https://github.com/kyma-project/kyma-environment-broker/issues/2409 

internal/broker/instance_update.go:250:1: cyclomatic complexity 60 of func `(*UpdateEndpoint).processUpdateParameters` is high (> 30) (gocyclo)
func (b *UpdateEndpoint) processUpdateParameters(ctx context.Context, instance *internal.Instance, details domain.UpdateDetails, lastProvisioningOperation *internal.ProvisioningOperation, asyncAllowed bool, ersContext internal.ERSContext, logger *slog.Logger) (domain.UpdateServiceSpec, error) {

Investigate and refactor.

PR #2899, opened 2026-02-20
`instance_update.go` refactor to reduce complexity

jaroslaw-pieszka:
**Description**

- Function extraction mostly

**Related issue(s)**

#2898

### DIFF
--- a/internal/broker/instance_update.go
+++ b/internal/broker/instance_update.go
@@ -390,10 +390,6 @@ func (b *UpdateEndpoint) validateMachineType(ctx context.Context, providerValues
 		}
 	}
 
-	if err != nil {
-		return err
-	}
-
 	if params.AdditionalWorkerNodePools != nil {
 		if err := b.validateAdditionalWorkerPoolsParams(details, params, ersContext, regionsSupportingMachine, instance, logger, providerValues, discoveredZones); err != nil {
 			return err


internal/broker/instance_update.go
```diff
@@ -474,6 +378,172 @@ func (b *UpdateEndpoint) processUpdateParameters(ctx context.Context, previousIn
...
+
+func (b *UpdateEndpoint) zeroFieldsForTrial(details domain.UpdateDetails, params internal.UpdatingParametersDTO) {
+	if details.PlanID == TrialPlanID {
+		params.MachineType = nil
+		params.AutoScalerMin = nil
+		params.AutoScalerMax = nil
+	}
+}
+
+func (b *UpdateEndpoint) getRegionsForMachineType(providerValues internal.ProviderValues, instance *internal.Instance, params internal.UpdatingParametersDTO) (internal.RegionsSupporter, error) {
```
KsaweryZietara: I think the function name should clearly indicate that it performs validation—something like `validateMachineTypeRegionSupport` would make that intent explicit. Also, is it really necessary for this function to return a `RegionsSupporter`?

jaroslaw-pieszka: After more substantial refactor this is no longer valid.

internal/broker/instance_update.go
```diff
@@ -474,6 +367,196 @@ func (b *UpdateEndpoint) processUpdateParameters(ctx context.Context, previousIn
...
+
+		if params.MachineType != nil && discoveredZones[*params.MachineType] < providerValues.ZonesCount {
+			message := fmt.Sprintf("In the %s, the %s machine type is not available in %v zones.", providerValues.Region, *params.MachineType, providerValues.ZonesCount)
+			return apiresponses.NewFailureResponse(fmt.Errorf("%s", message), http.StatusUnprocessableEntity, message)
+		}
+	}
+
+	if err != nil {
+		return err
+	}
```
KsaweryZietara: Is this error check actually necessary?

PR #2923, opened 2026-02-26
Assign multiple hyperscaler accounts for global accounts with a large number of clusters

MarekMichali:
**Description**

This feature allows the use of multiple credentialsBindings in one GA. This allows for provisioning a theoretically infinite number of clusters on one GA, as the previous limitation was based on how many clusters the hyperscaler allows to be provisioned in one account (credentialBinding).

The feature is disabled by default and provisioning follows the old path, using only one credentialBinding per GA without the possibility of using more credentialBindings. For data gathering we use KEB DB.

- Add a feature that allows to use multiple credentialBindings for one GA
- Add configuration option that allows to enable this feature only for selected GAs
- Add configuration option that allows to specify max cluster for one credentialBindings per provider and global
- Expand E2E test with a scenario where we check if new cluster are created on new credentialBindings
- Expand E2E test with a scenario where we check if new cluster are created on the least populated credentialBindings

**Related issue(s)**


docs/contributor/04-10-workflows.md
```diff
@@ -119,17 +119,19 @@ The [`run-keb-chart-integration-tests-reusable`](/.github/workflows/run-keb-char
...
-4. Creates required namespaces
-5. Installs dependencies required by the KEB chart
-6. Installs the KEB chart in the k3s cluster using `helm install`
-7. Waits for the KEB Pod to be ready
-8. Provisions an instance
-9. Updates the instance  
-10. Deprovisions the instance  
-11. Waits for all tests to finish
+1. Sets up a k3s cluster with required dependencies and installs the KEB chart
+2. Provisions the first instance and waits for it to complete successfully
```
IwonaLanger:
```suggestion
2. Starts provisioning the first instance and waits for it to complete successfully
```

IwonaLanger (approved):

PR #2965, opened 2026-03-10
ADR Abstract Machines Names

KsaweryZietara:
**Description**

- Introduce an ADR defining the naming conventions for Abstract Machines.

### DIFF
--- a/docs/contributor/adr/ADR-002-abstract-machines-names.md
+++ b/docs/contributor/adr/ADR-002-abstract-machines-names.md
@@ -916,7 +916,7 @@ In this model, inputs such as `m6i.large`, `m5.large` and `mi.large` can all be 
     }
     ```
 
-5. SRE migrates all existing Runtime CRs using Cluster Orchestrator, updating machine types (for example, `m6i.large` and `m5.large`) to the target version such as `m7i.large`. KEB configuration is already changed, so new worker pools are created with target version.
+5. SRE migrates all existing Runtime CRs using Cluster Orchestrator, updating machine types (for example, `m6i.large` and `m5.large`) to the target version, such as `m7i.large`. The KEB configuration is already changed, so new worker pools are created with the target version.
 6. After runtime migration is complete, update the KEB database so that instances no longer store explicit versioned values such as `m6i.large` or `m5.large`, and instead store only logical values such as `mi.large`.
 
 ## Decision


PK85 (commented):
Additional steps to check:

- [x] 1) This document needs to describe what will happen if e.g. m6i will be changed to m7i.
  - [x] 1.1) What will happen to all existing machines?
  - [x] 1.2) Such a change should run in the mainatance window, check with SRE
- [x] 2) If Version change, how consumption reporter will now to support that machine, check with Andreas Thaler. We need to have an e2e instruction for Operator/SRE what to do to update such machine type
- [x] 3) Add a section when old machine types will be still supported in the schema (e.g. m6i.xlarge, we do not want to break backward compatibility). Idea is that when abstract types will be added, an old types like m6i.xlarge must be as well resolved to the version that KEB supports for mi.xlarge. E.g Customer selected in the past m6i.xlarge, but KEB already used m7i.xlarge.
  - [x] 3.1) That means in the schema we will only have new machines, abstract ones. Check how clusters wil old machine types behaves in the BTP Cockpit in the  "View Parameters" and "Update Parameters" views
- [x] 4) We need a document describing what SRE / Operator needs to do, to change a machine Type version. Checking price, availability, etc

docs/contributor/adr/ADR-002-abstract-machines-names.md
```diff
@@ -0,0 +1,503 @@
...
+
+The resolution process is:
+1. Read the configured logical machine type (e.g., `general-2`).
+2. Look up the corresponding instance type in `machinesMapping`.
+3. Return the mapped instance type.
+
+This approach completely decouples the logical machine definition from the instance family, making upgrades or replacements easier by modifying only the mapping.
+
+# Comparison
+
```
a-thaler: I really like the full abstraction as it makes the selection consistent across the provider, however, for GPU we will for sure support more types in future. Also think of ARM machines which we potentially support in future also for general purpose. So we need to stay flexible in supporting more types of a category

docs/contributor/adr/ADR-002-abstract-machines-names.md
```diff
@@ -0,0 +1,503 @@
...
+To simplify upgrades between instance generations, the configuration can be partially abstracted.
+Instead of referencing full instance family names, a logical machine type is used. The actual family is then resolved through a mapping.
+
+## AWS
+
+```yaml
+providersConfiguration:
+  aws:
+    machines:
+      mi.large: mi.large (2vCPU, 8GB RAM)
```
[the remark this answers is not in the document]
KsaweryZietara: Yes, we could deprecate **m5** and keep only **m6i** within the `m` family, but in this case we can’t support both at the same time.

docs/contributor/adr/ADR-002-abstract-machines-names.md
```diff
@@ -0,0 +1,951 @@
...
+    ],
+    "type": "string"
+  }
+}
+```
+
+The JSON schema requires that the `machineType` value must be one of the entries defined in the enum list.
+It does not allow values outside of this predefined set.
+
+If abstract machine types are introduced (e.g., `m.large`), the existing machine types must remain in the schema to maintain backward compatibility.
```
IwonaLanger:
```suggestion
If abstract machine types are introduced (for example, `m.large`), the existing machine types must remain in the schema to maintain backward compatibility.
```

PR #2978, opened 2026-03-13
Take into account CB with tenantName when no CB in DB

MarekMichali:
**Description**

The current implementation of multi account only allows the use of CBs that are in the KEB DB when a CB with a tenantName is returned by HAP or allows claiming a new fresh CB without a tenantName. For development purposes, we have CBs with the label "internal" that prevent marking a CB as dirty when the last instance using this CB is deprovisioned. After deprovisioning like this, the KEB DB contains no instances that use a CB with a tenantName returned by HAP. This PR allows the use of such CBs that have no instances in the DB but have a tenantName label.

- Support in multi account mode CBs that have a tenantName label but no instances in DB
- Fix typo in tests
- More verbose logs

**Related issue(s)**

### DIFF
--- a/internal/process/provisioning/resolve_credentials_binding.go
+++ b/internal/process/provisioning/resolve_credentials_binding.go
@@ -212,21 +212,8 @@ func (s *ResolveCredentialsBindingStep) resolveWithMultiAccountSupport(operation
 			return "", fmt.Errorf("while getting instance counts per binding: %w", err)
 		}
 
-		// Find the most populated binding that is still below the limit.
-		// Bindings with tenantName label but 0 instances in KEB DB are also considered
-		selectedBinding := ""
-		selectedBindingCount := -1
-		for _, name := range bindingNames {
-			count := instancesPerBinding[name]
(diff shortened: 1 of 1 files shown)


internal/process/provisioning/resolve_credentials_binding.go
```diff
@@ -201,19 +201,33 @@ func (s *ResolveCredentialsBindingStep) resolveWithMultiAccountSupport(operation
...
-			return "", fmt.Errorf("while getting best credentials binding: %w", err)
+			return "", fmt.Errorf("while getting instance counts per binding: %w", err)
 		}
-		if bestBindingName != "" {
-			log.Info(fmt.Sprintf("selected credentials binding %s with %d instances (below limit %d)", bestBindingName, count, hyperscalerAccountLimit))
-			return bestBindingName, nil
+
+		// Find the most populated binding that is still below the limit.
+		// Bindings with tenantName label but 0 instances in KEB DB are also considered
+		selectedBinding := ""
```
jaroslaw-pieszka: consider extracting as separate function/method.

PR #3053, opened 2026-04-08
Operation blocklist MVP

jaroslaw-pieszka:
**Description**

- Blocking provision, update, planUpgrade per plan

**Related issue(s)**


jaroslaw-pieszka:
More probably we will use global operation blocklist.

PR #3078, opened 2026-04-16
MVP Global blocklisted operations

jaroslaw-pieszka:
**Description**

- Possible blocking for four operations with specified plan(s) or blocking for all plans

**Related issue(s)**


cmd/broker/main.go
```diff
@@ -541,6 +543,13 @@ func createAPI(router *httputil.Router, schemaService *broker.SchemaService, ser
 	fatalOnError(err, logs)
 	logs.Info(fmt.Sprintf("Number of subaccountIds with unlimited quota: %d", len(quotaWhitelistedSubaccountIds)))
 
+	var operationBlocklist blocklist.OperationBlocklist
+	if cfg.OperationBlocklistFilePath != "" {
+		operationBlocklist, err = blocklist.ReadFromFile(cfg.OperationBlocklistFilePath)
+		fatalOnError(err, logs)
+	}
+	operationBlocklist = operationBlocklist.WithPlanValidator(broker.AvailablePlans)
```
KsaweryZietara, line 551: Above, we check whether `cfg.OperationBlocklistFilePath` is non-empty. Will `WithPlanValidator` still work if it is empty?

KsaweryZietara:
I think we should add end-to-end tests for this in `instance_create_test.go`, `instance_update_test.go`, and `instance_deprovision_test.go`.

KsaweryZietara:
Do we really need to make the message configurable? It adds a lot of complexity to the code.

From what I understand, in most cases—if not all—the message will follow the same pattern: 
```
“{operation type} is blocked for the {plan} plan.” 
```
If that’s true, why not hardcode the message template, as we do in other cases, and make only the list of plans configurable?

That would significantly reduce the complexity.

cmd/broker/main.go
jaroslaw-pieszka: Will check.

jaroslaw-pieszka:
> I think we should add end-to-end tests for this in `instance_create_test.go`, `instance_update_test.go`, and `instance_deprovision_test.go`.

Will add.

jaroslaw-pieszka:
> Do we really need to make the message configurable? It adds a lot of complexity to the code.
> 
> From what I understand, in most cases—if not all—the message will follow the same pattern:
> 
> ```
> “{operation type} is blocked for the {plan} plan.” 
> ```
> 
> If that’s true, why not hardcode the message template, as we do in other cases, and make only the list of plans configurable?
> 
> That would significantly reduce the complexity.

Being discussed in private channels.

jaroslaw-pieszka:
Superseded by #3092 which includes the same implementation plus end-to-end tests in instance_create_test.go, instance_update_test.go, and instance_deprovision_test.go.

PR #3102, opened 2026-04-20
OperationBlocklist validation hardened

jaroslaw-pieszka:
## Summary

- Empty string `''` rule is now a no-op instead of a parse error
- Empty message `'""'` fails loading with a parse error
- Unknown plan names (e.g. typo `trail` instead of `trial`) are rejected at startup via `WithPlanValidator`, which now returns an error
- Empty file or whitespace-only file is treated as a no-op blocklist
- All cases covered by tests

## Test plan

- [ ] `go test ./internal/blocklist/...` — all 29 tests pass
- [ ] `go build ./cmd/broker/...` — broker compiles


jaroslaw-pieszka:
Superseded by a conflict-free rebase on main

PR #3128, opened 2026-04-27
Update default machine type for AWS Free and Trial

KsaweryZietara:
**Description**

This pull request updates the default AWS machine types for Free and Trial plans:
- when `UseSmallerMachineTypes` is set to true, the default machine type is `mi.large`.
- when `UseSmallerMachineTypes` is set to false, the default machine type is `mi.xlarge`.

**Related issue(s)**
See also [#9240](https://github.tools.sap/kyma/backlog/issues/9240)


MarekMichali (approved):

cmd/broker/provisioning_test.go
```diff
@@ -1474,7 +1474,7 @@ func TestProvisioning_ClusterParameters(t *testing.T) {
 			expectedZonesCount:           ptr.Integer(1),
 			expectedMinimalNumberOfNodes: 3,
 			expectedMaximumNumberOfNodes: 20,
-			expectedMachineType:          provider.DefaultAWSMachineType,
```
PK85, line 1487: why hardcoding? Maybe we should use new DefaultAWSFreeTrialMachineType

KsaweryZietara: The test name is `Production AWS`, so we should not use `DefaultAWSFreeTrialMachineType`, which is set to `mi.xlarge`. That value is intended for free and trial plans, and using it here would be misleading.

PR #3178, opened 2026-05-11
Add integration test for autoScalerMin changes reflected in runtimes endpoint

jaroslaw-pieszka:
## Summary

- Adds `TestUpdateAutoScalerMinReflectedInRuntimes` to `cmd/broker/update_test.go`
- Test provisions an instance with `autoScalerMin=4` via PUT, then updates it to 5 and 6 via two PATCH calls, then calls GET `/runtimes` and asserts the correct values appear on the instance parameters, provisioning operation, and both update operations
- Verifies end-to-end that `autoScalerMin` changes are correctly stored per-operation and returned by the runtimes endpoint

## Test Plan
- [ ] `go test ./cmd/broker/... -run TestUpdateAutoScalerMinReflectedInRuntimes` passes


piotrek1459 (approved):
/lgtm

Issue #3043, opened 2026-04-07
KEB Observability — Parameter Usage Analytics


PK85:
# KEB Observability — Parameter Usage Analytics UI

## User Story

**As a** KEB platform operator / SRE,
**I want** a self-contained analytics UI backed by a new KEB API endpoint showing which provisioning and update parameters were used across all Kyma instances — ranked from most to least used —
**So that** I can understand adoption patterns, identify deprecated or underused parameters, and guide future API evolution decisions.

---

## Background & Context

When a Kyma instance is provisioned or updated, a rich set of parameters is persisted as JSON in the `instances.provisioning_parameters` and `operations.data` columns in PostgreSQL. These include cloud provider, region, machine type, OIDC config, autoscaler settings, additional worker node pools, modules, networking, and more.

Currently there is no aggregated view of parameter usage. This story delivers that via a new `GET /runtimes/analytics` KEB endpoint and a lightweight `keb-analytics` UI binary — both living in the KEB repository and Helm chart — replacing the previously proposed Plutono dashboard approach.

---

### AC0 — Extensibility
- [x] The solution must support any new provisioning or update parameter seamlessly. If a new parameter (e.g. `isDay: true/false`) is added, all UI views must display its adoption out of the box without code changes.

### AC1 — General Parameter Usage View
- [x] A ranked bar chart showing the percentage of instances that had each parameter explicitly set (non-null / *(stretch)* non-default), e.g.:
  - `region` set: 94%
  - `machineType` set: 78%
  - `autoScalerMin` set: 61%
  - `oidc` set: 55%
  - `modules` set: 43%
  - `additionalWorkerNodePools` set: 12%
  - `networking.pods` set: 8%
  - `ingressFiltering` set: 3%
- [x] Filterable by **plan** (aws, azure, gcp, sap-converged-cloud, trial, free, own-cluster)
- [x] Filterable by **region**
- [x] Filterable by time range (last 7d / 30d / 90d / all time)

### AC2 — Provisioning Parameter Usage View
- [x] Same ranked bar chart scoped to **provisioning operations** (`instances.provisioning_parameters`)
- [x] Filterable by **plan**, **region**, and time range

### AC3 — Update Parameter Usage View
- [x] Ranked bar chart scoped to **update operations** (`operations.data` / `UpdatingParametersDTO`), covering:
  - `autoScalerMin` / `autoScalerMax`
  - `machineType`
  - `oidc`
  - `runtimeAdministrators`
[...]

PR #3198, opened 2026-05-14
Make Resource State Retry Interval Configurable and Speed Up Local Seeding

jaroslaw-pieszka:
**Description**

- Add APP_STEP_TIMEOUTS_RESOURCE_STATE_RETRY_INTERVAL env var (default 10s) to StepTimeoutsConfig, replacing the hardcoded constant
- Wire the new config field into the provisioning and update step pipelines
- Add the field to values.yaml and deployment.yaml for easy override in local development
- Rewrite seed_analytics.py to fire all provision requests concurrently using ThreadPoolExecutor; add --provision-workers flag (default 50)
- Patch Runtime CRs concurrently in seed_analytics.py
- Fix suite_test.go to set ResourceStateRetryInterval=10ms to prevent zero-value default from breaking tests

**Related issue(s)**
See also #3043


IwonaLanger (approved):

jaroslaw-pieszka:
Outdated.

PR #3207, opened 2026-05-15
Improve keb-analytics UI: axis labels, top-10 distribution, linear trend scale

jaroslaw-pieszka:
## Summary

- Add X/Y axis titles to Combined, Provisioning, Update, and Trends charts
- Cap Value Distribution at top 10 buckets; collapse the rest into an "others (N more values)" bar
- Sync Value Distribution dropdown with "Include not provided/null" checkbox: when checked, parameters with zero set values (100% not provided) also appear in the dropdown
- Render Trends X-axis as a linear scale over the full selected date range with day+month tick labels ("Jan 05")
- Rewrite Trends tab description in plain language
- Clarify "Total active instances" label with a note that it reflects current filter selection
- Include `behaviorCount` fields (e.g. `additionalWorkerNodePools`, `administrators`) in Value Distribution — previously excluded, causing a mismatch with Combined/Provisioning tabs
- Add tests confirming combined set-count bounds relative to provisioning and update distinct instance counts
- Add `TestBuildDistributions_ParamSetConsistency` to guard that every provisioning/combined parameter appears in distributions

Part of #3043


cmd/keb-analytics/static/index.html
```diff
@@ -2,8 +2,11 @@
 <html lang="en">
 <head>
   <meta charset="UTF-8">
-  <title>KEB Analytics</title>
-  <script src="https://cdn.jsdelivr.net/npm/chart.js@4"></script>
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <title>KEB Parameter Usage Analytics</title>
+  <script src="https://cdn.jsdelivr.net/npm/chart.js@4.4.9/dist/chart.umd.min.js"
```
piotrek1459, line 7: https://github.com/chartjs/Chart.js
this was released 1 year ago, so probably was not affected by the last supply chain attack, but I couldn't find the final proof. https://security.snyk.io/TanStack-npm-Supply-Chain-Compromise-May-2026
here there is no mention of that package so I think it is safe.

jaroslaw-pieszka: For a direct npm dependency, Chart.js currently shows no evidence of being part of the May 2026 TanStack/“Mini Shai‑Hulud” or Axios npm supply‑chain compromises and is considered safe to use when installed from the official `chart.js` package at a known‑good released version (such as 4.5.1) with a locked dependency tree and routine SCA scanning in place. [orca](https://orca.security/resources/blog/axios-npm-supply-chain-attack-remediation/)

internal/analytics/aggregator_test.go
```diff
@@ -203,11 +205,198 @@ func TestBuildDistributions_IncludesRegion(t *testing.T) {
...
+		if d.Parameter == "additionalWorkerNodePools" {
+			assert.Equal(t, 1, d.Values["2"], "i1 has 2 pools")
+			assert.Equal(t, 1, d.Values["1"], "i2 has 1 pool")
+			found = true
+		}
+	}
+	assert.True(t, found, "additionalWorkerNodePools should appear in distributions")
+}
+
+// TestBuildDistributions_ParamSetConsistency confirms that every parameter appearing
```
piotrek1459, line 365: TestBuildDistributions_ParamSetConsistency — the test passes, but the invariant stated in the comment is not actually guaranteed by the system, and the test data accidentally masks this.

The assertion "combined param %q must appear in distributions" holds here only because i1 provisioned with gvisor, so BuildDistributions(prov) includes it. But AggregateCombined also picks up parameters that were only ever set via an update, never at provisioning time — and BuildDistributions takes only provisioning params, so those would be absent from distributions.

The following data would fail the test:


prov := []ProvisioningParamsWithID{
    // neither instance provisions with gvisor
    {InstanceID: "i1", Params: internal.ProvisioningParameters{Parameters: pkg.ProvisioningParametersDTO{
        Region: &region,
    }}},
    {InstanceID: "i2", Params: internal.ProvisioningParameters{Parameters: pkg.ProvisioningParametersDTO{
        Region: &region,
    }}},
}
upd := []UpdateParamsWithID{
    // only i2 sets gvisor — via update
    {InstanceID: "i2", Params: internal.UpdatingParametersDTO{Gvisor: &pkg.GvisorDTO{Enabled: true}}},
}
Here combined.Parameters contains gvisor (count: 1, from i2's update), but BuildDistributions(prov) does not — no provisioning operation ever set it. The assertion "combined param "gvisor" must appear in distributions" fails.

The JS layer handles this correctly via populateDistParam (when "include null" is checked, combined params missing from distributions are added to the dropdown with 100% not-provided). But the test comment overpromises — it should be scoped to provisioning params only, or the data should be adjusted to document the actual invariant being verified.

piotrek1459:
#3213 review enhancements handled here.

Issue #3039, opened 2026-04-07
Worker Pool Annotations and Labels Configuration


PK85:
## Summary

Enable customers to configure **labels** and **annotations** per worker pool node via the BTP Cockpit UI, and pass this data into the RuntimeCR.

## Background

This is the second release of the worker pool configuration feature (following the completed [taints](https://github.com/kyma-project/kyma-environment-broker/issues/2024) release). Labels and annotations allow customers to organize and attach metadata to worker pool nodes, enabling workload scheduling, observability tooling, and cluster management integrations.

Progress is currently blocked on BTP UI team implementing the necessary schema changes in BTP Cockpit (expected Q2 2026).

### BTP Cockpit UI
- [ ] KEB UI supports defining key-value entries per worker pool for **Labels**
- [ ] KEB UI supports defining key-value entries per worker pool for **Annotations**
- [ ] Align with BTP UI team on the input format:
  - Solution:
    ```json
    "labels": [
      "asdas": "sds",
      "asdasasd": "ass"
    ]
    ```
  > **Decision needed:** Confirm chosen format with BTP UI team before implementation starts. Option B is preferred for consistency with the taints implementation.

### RuntimeCR Integration
- [ ] Labels and annotations from UI are correctly passed into the RuntimeCR
- [ ] Align with KIM team (`@kyma-project/framefrog`) on the exact RuntimeCR field structure

### Testing
- [ ] Verify whether KIM team already covers this in their e2e test suite
- [ ] Extend SKR e2e tests to cover labels and annotations on worker pools

### Documentation
- [ ] Update internal documentation
- [ ] Update SAP Help Portal
- [ ] Release Note (RN) prepared

### ERS Rollout
- [ ] Update ERS and validate e2e on:
  - [ ] dev
  - [ ] stage
  - [ ] prod

### Demo
- [ ] Prepare e2e demo for Iteration Review

## Dependencies / Blockers

- **BTP UI team** must implement BTP Cockpit UI schema update (tracked in [SAPBTPCFS-26528](https://jira.tools.sap/browse/SAPBTPCFS-26528))
- BTP UI team confirmed Q2 2026 delivery — ticket stays blocked until then

## Status History

| Date       | Update |
|------------|--------|
| 2025-09-04 | Blocked by BTP UI. Ticket SAPBTPCFS-26528 raised for BTP Cockpit schema update |
| 2025-09-30 | BTP UI team not yet responding |
[...]

PR #3239, opened 2026-06-01
Labels and annotations 

piotrek1459:
**Description**

- Added support for labels and annotations on additional worker node pools — users can now set key-value metadata per pool via provisioning and update requests
- Labels and annotations are passed through to the Gardener Worker struct on the RuntimeCR; omitting or providing an empty map removes them (full-replacement semantics, same as taints)
- Validation rejects empty keys; schema exposes both fields in the UI with a keyvalue control type
- Tests: unit (validation, provider mapping), integration (add/update/remove/mixed scenarios), docs updated

**Related issue(s)**

https://github.com/kyma-project/kyma-environment-broker/issues/3039


cmd/broker/broker_suite_test.go
```diff
@@ -1073,6 +1073,28 @@ func (s *BrokerSuiteTest) assertAdditionalWorkerTaints(t *testing.T, provider im
 	assert.ElementsMatch(t, expectedTaints, worker.Taints)
 }
 
+func (s *BrokerSuiteTest) assertAdditionalWorkerLabels(t *testing.T, provider imv1.Provider, name string, expectedLabels map[string]string) {
```
jaroslaw-pieszka, line 1080: I am just wondering if we can avoid duplication of code.
We iterate over all AdditionalWorkers to find particular one and then we check some assertion.
Could be that this is overkill but let's do some mental experiment.
assertForAdditionalWorkingNodePool(t, provider, name, f())

common/runtime/model.go
```diff
@@ -642,6 +644,24 @@ func (a AdditionalWorkerNodePool) ValidateTaints(taints []TaintDTO, poolName str
 	return nil
 }
 
+func (a AdditionalWorkerNodePool) ValidateLabels(labels map[string]string, poolName string) error {
```
jaroslaw-pieszka: Code duplication. The only difference is the error message.

cmd/broker/update_test.go
```diff
@@ -3933,6 +3933,205 @@ func TestUpdateAdditionalWorkerNodePools(t *testing.T) {
...
+				"parameters": {
+					"name": "testing-cluster",
+					"region": "eu-central-1"
+				}
+			}`)
+		defer func() { _ = resp.Body.Close() }()
+		opID := suite.DecodeOperationID(resp)
+		suite.waitForRuntimeAndMakeItReady(opID)
+		suite.WaitForOperationState(opID, domain.Succeeded)
+
```
jaroslaw-pieszka, line 3964: Assert no workers pools (hence no labels, no annotations) before update.

cmd/broker/update_test.go
```diff
@@ -3933,6 +3933,205 @@ func TestUpdateAdditionalWorkerNodePools(t *testing.T) {
...
+		// then
+		suite.WaitForOperationState(upgradeOperationID, domain.Succeeded)
+		runtime := suite.GetRuntimeResourceByInstanceID(iid)
+		assert.Len(t, *runtime.Spec.Shoot.Provider.AdditionalWorkers, 1)
+		suite.assertAdditionalWorkerLabels(t, runtime.Spec.Shoot.Provider, "labeled-pool", map[string]string{"env": "prod", "team": "platform"})
+		suite.assertAdditionalWorkerAnnotations(t, runtime.Spec.Shoot.Provider, "labeled-pool", map[string]string{"note": "test"})
+	})
+
+	t.Run("should update labels on existing additional worker node pool", func(t *testing.T) {
```
jaroslaw-pieszka, line 4000: What about scenario with update on not existing pool?
What about mixed update (not existing with existing)?

cmd/broker/update_test.go
```diff
@@ -3933,6 +3933,205 @@ func TestUpdateAdditionalWorkerNodePools(t *testing.T) {
...
+							"labels": {"env": "prod"}
+						}
+					]
+				}
+			}`)
+		defer func() { _ = resp.Body.Close() }()
+		opID := suite.DecodeOperationID(resp)
+		suite.waitForRuntimeAndMakeItReady(opID)
+		suite.WaitForOperationState(opID, domain.Succeeded)
+
```
jaroslaw-pieszka, line 4259: Assert before update (pre-condition).

internal/workers/provider_test.go
```diff
@@ -323,6 +323,88 @@ aws:
 		assert.Equal(t, corev1.Taint{Key: "dedicated", Value: "ml", Effect: corev1.TaintEffectPreferNoSchedule}, workers[0].Taints[1])
 	})
 
+	t.Run("should map labels and annotations to gardener worker", func(t *testing.T) {
```
jaroslaw-pieszka, line 326: Maps with empty key?
You have considered such a scenario, is it tested?

common/runtime/model.go
piotrek1459: Good catch. Added a private validateKeyValueMap helper — both methods now delegate to it.

cmd/broker/update_test.go
piotrek1459: Added assertion that no additional workers exist after provisioning before the update is sent.

cmd/broker/update_test.go
piotrek1459: Added two new test cases: one for updating a non-existing pool name (full replacement), and one for a mixed update (existing pool updated + new pool added).

cmd/broker/update_test.go
piotrek1459: Added assertion that labels are set after provisioning before the removal update is sent.

internal/workers/provider_test.go
piotrek1459: The empty-key scenario is validated in checkLabelsConfiguration/checkAnnotationsConfiguration (not in CreateAdditionalWorkers), so the right place is instance_create_test.go. Added three new table entries covering label with empty key, annotation with empty key, and valid labels+annotations.

cmd/broker/broker_suite_test.go
piotrek1459: I propose extracting the worker lookup into a single assertForAdditionalWorkerNodePool helper that accepts a func(*v1beta1.Worker) callback — each specific assert then just passes a closure with the actual assertion. This removes the repeated find-loop from every method and makes adding new assertions a 3-line change.

cmd/broker/broker_suite_test.go
```diff
@@ -1060,29 +1061,42 @@ func (s *BrokerSuiteTest) assertAdditionalWorkerIsCreated(t *testing.T, provider
...
 
-func (s *BrokerSuiteTest) assertAdditionalWorkerTaints(t *testing.T, provider imv1.Provider, name string, expectedTaints []corev1.Taint) {
-	var worker *v1beta1.Worker
-	for _, additionalWorker := range *provider.AdditionalWorkers {
-		if additionalWorker.Name == name {
-			worker = &additionalWorker
+func (s *BrokerSuiteTest) assertForAdditionalWorkerNodePool(t *testing.T, provider imv1.Provider, name string, f func(*v1beta1.Worker)) {
+	for _, w := range *provider.AdditionalWorkers {
+		if w.Name == name {
+			wCopy := w
```
jaroslaw-pieszka: Do we need the copy?

piotrek1459: No, we don't need it.

internal/broker/instance_create.go
```diff
@@ -780,7 +798,45 @@ func checkTaintsConfiguration(additionalWorkerNodePools []pkg.AdditionalWorkerNo
...
+			errors = append(errors, err.Error())
+		}
+	}
+
+	if len(errors) == 0 {
+		return nil
+	}
+
+	message := additionalWorkerPoolsValidationIssuesMsg
+	message = message + strings.Join(errors, "; ")
```
jaroslaw-pieszka: This code is repeated, we should extract and reuse.

piotrek1459: done. Now this part is delegated.

jaroslaw-pieszka (dismissed):

docs/user/04-40-additional-worker-node-pools.md
```diff
@@ -157,6 +157,176 @@ The update operation overwrites the additional worker node pools with the list p
     }
     ```
 
+## Labels
+
+Each additional worker node pool supports an optional **labels** map. Labels are key-value pairs attached to the worker nodes, which you can use to identify, filter, and organize them.
```
IwonaLanger:
```suggestion
Each additional worker node pool supports an optional **labels** map. Labels are key-value pairs attached to worker nodes that you can use to identify, filter, and organize them.
```

docs/user/04-40-additional-worker-node-pools.md
```diff
@@ -157,6 +157,176 @@ The update operation overwrites the additional worker node pools with the list p
...
+
+Each label entry has the following properties:
+
+| Property | Required | Description |
+|----------|----------|-------------|
+| **key** | Yes | Any non-empty string |
+| **value** | No | Any string |
```
IwonaLanger: We use the bold font for actual parameter/key/variable names. From what I see in the example, 'key' and 'value' are more like placeholders for whatever is expressed as a string. Do I understand it correctly?

docs/user/04-40-additional-worker-node-pools.md
```diff
@@ -157,6 +157,176 @@ The update operation overwrites the additional worker node pools with the list p
...
+## Annotations
+
+Each additional worker node pool supports an optional **annotations** map. Annotations are key-value pairs used to attach arbitrary non-identifying metadata to worker nodes, such as tooling configuration or operational notes.
+
+Each annotation entry has the following properties:
+
+| Property | Required | Description |
+|----------|----------|-------------|
+| **key** | Yes | Any non-empty string |
+| **value** | No | Any string |
```
IwonaLanger: Please see the comment above.

docs/user/04-40-additional-worker-node-pools.md
piotrek1459: sure, done.

docs/user/04-40-additional-worker-node-pools.md
piotrek1459: done.

IwonaLanger (approved):

jaroslaw-pieszka (approved):

PR #3241, opened 2026-06-01
Add Analytics Tests, Fix Trend X-axis, and Fix Trends Tab for Time-Range Filter

jaroslaw-pieszka:
**Description**

**Tests (`internal/analytics/aggregator_test.go`)**
- Add `TestWalkFields_AggregatesAcrossMultipleInstances` — verifies count accumulation across multiple calls
- Strengthen `TestAggregateProvisioning_RanksParameters` and `TestAggregateUpdates_CountsSetFields`: assert correct `Total`, use `CountFor`, verify sort order
- Add 6 `AggregateCombined` tests: provisioning-only, update-only, union, deduplication, total equals provisioning count, sort order
- Add 2 `AggregateCombined` tests for decreasable parameters: nullification by update does not remove instance from Combined set (contrast with `BuildTrend`)
- Add 7 `BuildTrend` tests: param set/not-set at provision, added/removed by update, non-updatable param preserved, Total accumulation, multi-instance multi-day
- Add tests for `FilterByPlan`, `FilterByRegion`, `BuildPlanRegionIndex`

**Comment fixes (`aggregator.go`, `model.go`)**
- Fix `walkFields` doc comment to describe all value-transformation behaviors
- Fix `TrendPoint.Total` and `BuildTrend` comments: Total is cumulative provisioned count, never decremented

**Bug fixes (`cmd/keb-analytics/`)**
- Fix Trends tab showing nothing when a time-range filter (7d/30d/90d) is active: op events are now always fetched with an unbounded range so `BuildTrend` has full history; the period filter applies to counts only
- Fix trend chart X-axis: replace category scale with numeric timestamp scale so points are spaced by actual calendar distance; ticks formatted as `Mon DD` with `maxTicksLimit: 12`

**Related issue(s)**
See also #3043


internal/analytics/aggregator_test.go
```diff
@@ -214,6 +230,420 @@ func TestBuildDistributions_IncludesRegion(t *testing.T) {
...
+	trend := BuildTrend(events, "machineType")
+	require.Len(t, trend.Points, 2)
+	assert.Equal(t, 0, trend.Points[0].Count) // day 1: no param yet
+	assert.Equal(t, 1, trend.Points[1].Count) // day 2: update sets it → +1
+}
+
+func TestBuildTrend_ParamRemovedByUpdate(t *testing.T) {
+	// provision with param, then update clears it (machineType is NOT in updatingFieldConfig,
+	// so its absence in an update op means "state unchanged" — use gvisor which IS updatable)
+	provWithGvisor := func(instanceID, day string) OpEvent {
```
KsaweryZietara: `provWithGvisor` here duplicates `provEventGvisor` defined above — can just call that directly.

KsaweryZietara (approved):

Issue #3152, opened 2026-05-06
Configuration for KEB allow to provide a keys with null values and it doesnt work with Landscaper


PK85:
**Description**

Landscaper uses strategic merge patch semantics when processing data objects. Null values are interpreted as "delete this key" rather than "set this key to null". This is what we saw in the kcp-kyma-environment-broker configmap too.

This config:
```
providersConfiguration:
    aws:
        regionsSupportingMachine:
            g4dn:
                us-gov-east-1:
                us-gov-west-1:
            g6:
```
becomes:
```
providersConfiguration:
    aws:
        regionsSupportingMachine:
            g4dn: { }
``` 

**AC**

- [x] 1) check where in our configs we allow to leave keys with null values
  - [x] 1.1) PK: I checked that [THIS](https://github.com/kyma-project/kyma-environment-broker/issues/3152#issuecomment-4388126193) works (I tested that with zonesDiscovery: true and zonesDiscovery: false)
  - [x] 1.2) PK: I checked that [THIS](https://github.com/kyma-project/kyma-environment-broker/issues/3152#issuecomment-4388265317) works. 
- [x] 2) for regionsSupportingMachine https://github.com/kyma-project/kyma-environment-broker/pull/3264
  - [x] 2.1) Add missing documentation in the [KEB](https://github.com/kyma-project/kyma-environment-broker/blob/main/docs/contributor/03-50-regions-supporting-machine.md) and Restricted Markets about 1.1) 
  - [x] 2.2) Add missing tests in the KEB and Restricted Markets about 1.1) including zonesDiscovery: true (mock real call, but test this scenario too)
  - [x] 2.3) Add missing documentation in the [KEB](https://github.com/kyma-project/kyma-environment-broker/blob/main/docs/contributor/03-50-regions-supporting-machine.md) and Restricted Markets about 1.2) 
  - [x] 2.4) Add missing tests in the KEB and Restricted Markets about 1.3) including zonesDiscovery: true (mock real call, but test this scenario too)
- [x] 3) Looks code already handles that, but double check. This is an enhancement(backward compatible), release as a patch with updated dosc and tests.
- [x] 4)  Consider changing default to true https://github.com/kyma-project/kyma-environment-broker/pull/3265

    ```
    # If true, enables provisioning of clusters with nodes distributed across multiple availability zones.
      multiZoneCluster: "false"
    ```

PK85:
I tested a settings:
input:
```
[...]

PR #3264, opened 2026-06-08
Add Landscaper Compatibility documentation

KsaweryZietara:
**Description**

- Fix error message in `checkUnsupportedMachines` to handle machine types not supported in any region (instead of joining an empty slice)
- Document Landscaper compatibility for `regionsSupportingMachine` — explain why bare null keys are dropped under strategic merge patch and how to use `[]` and `{}` instead
- Add the test cases described in the issue

**Related issue(s)**
See also #3152

### DIFF
--- a/docs/contributor/03-50-regions-supporting-machine.md
+++ b/docs/contributor/03-50-regions-supporting-machine.md
@@ -72,7 +72,7 @@ regionsSupportingMachine:
 
 ### Machine type not available in any region
 
-Use an **empty map `{}`** as the value for the machine family:
+Use an empty map `{}` as the value for the machine family:
 
 ```yaml
 regionsSupportingMachine:
--- a/docs/contributor/03-50-regions-supporting-machine.md
+++ b/docs/contributor/03-50-regions-supporting-machine.md
@@ -79,7 +79,7 @@ regionsSupportingMachine:
     g6: {}
 ```
 
-This makes the machine type unavailable in every region. Users attempting to provision in any region will receive an error similar to:
+This makes the machine type unavailable in every region. Users attempting to provision in any region receive an error similar to:
 
 ```
  In the region eu-central-1, the following machine types are not available: g6.xlarge (used in: worker-1), not supported in any region
--- a/docs/contributor/03-50-regions-supporting-machine.md
+++ b/docs/contributor/03-50-regions-supporting-machine.md
@@ -51,6 +51,9 @@ regionsSupportingMachine:
         eu-central-1:
         eu-west-2:
     g6:
+```
+
+results in:
 
 ```yaml
 regionsSupportingMachine:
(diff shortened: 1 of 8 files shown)


docs/contributor/03-50-regions-supporting-machine.md
```diff
@@ -29,14 +29,61 @@ providersConfiguration:
...
+
+```yaml
+regionsSupportingMachine:
+    g4dn:
+        eu-central-1:
+        eu-west-2:
+    g6:
+```
+
+are silently dropped by Landscaper, resulting in:
```
IwonaLanger:
```suggestion
Landscaper uses strategic merge patch semantics when applying ConfigMaps. Under these semantics, a key with a null value is interpreted as "delete this key" rather than "set this key to null". This means that region entries written without a value are silently dropped by Landscaper. Therefore, the following setting:

```yaml
regionsSupportingMachine:
    g4dn:
        eu-central-1:
        eu-west-2:
    g6:
```

results in:
```

File: docs/contributor/03-50-regions-supporting-machine.md [hunk 2]
```diff
@@ -29,14 +29,61 @@ providersConfiguration:
...
+
+```yaml
+regionsSupportingMachine:
+    g4dn: {}
+```
+
+### Region supported in all its zones (no zone restriction)
+
+Use an **empty list `[]`** instead of a bare key (null value):
```
IwonaLanger:
```suggestion
Use an empty list `[]` instead of a bare key (null value):
```

docs/contributor/03-50-regions-supporting-machine.md
```diff
@@ -29,14 +29,61 @@ providersConfiguration:
...
+    g4dn:
+        eu-central-1: []
+        eu-west-2: []
+```
+
+`eu-central-1: []` means the region is supported but zone selection falls back to the Kyma worker node pool zones. This is functionally equivalent to a bare key when processed by KEB, but survives the Landscaper patch without being dropped.
+
+### Machine type not available in any region
+
+Use an **empty map `{}`** as the value for the machine family:
```
IwonaLanger:
```suggestion
Use an empty map `{}` as the value for the machine family:
```

docs/contributor/03-50-regions-supporting-machine.md
```diff
@@ -29,14 +29,61 @@ providersConfiguration:
...
+
+```yaml
+regionsSupportingMachine:
+    g6: {}
+```
+
+This makes the machine type unavailable in every region. Users attempting to provision in any region will receive an error similar to:
```
IwonaLanger:
```suggestion
This makes the machine type unavailable in every region. Users attempting to provision in any region receive an error similar to:
```

jaroslaw-pieszka (dismissed):

IwonaLanger (approved):

jaroslaw-pieszka (approved):

PR #3278, opened 2026-06-09
fix(metrics): rename "shoot_name" metric label to "shoot"

amineelarabisap:
## Summary

Rename the `shoot_name` metric label to `shoot` in the `kcp_keb_v2_operation_result` metric. This aligns with the standard label name used by other KCP components and removes the need for `label_replace` workarounds in alerting rules.

## Changes

- `internal/metrics/metrics.go`: label key `"shoot_name"` → `"shoot"` in `GetLabels()`
- `internal/metrics/operations_result.go`: GaugeVec label declaration updated
- `internal/metrics/metrics_test.go`: assertions updated

## After deployment

Alert rules in `management-plane-charts` branch `chart/keb-sap` (`files/misc/prometheus-rules.yaml`) that use `label_replace(..., "shoot_name", ...)` or reference `shoot_name` in `sum by (...)` clauses can be simplified to use `shoot` directly.

Ref: https://github.tools.sap/kyma/backlog/issues/9360


amineelarabisap:
CI requires a `kind/bug` label. Could a maintainer add it? The `govulncheck` failure seems to be a pre-existing Docker dependency vulnerability and unrelated to this change.

MarekMichali (approved):

PR #3301, opened 2026-06-12
Extract parseGAList Helper to Reduce parseRule Complexity

jaroslaw-pieszka:
**Description**

Changes proposed in this pull request:

- Extract `parseGAList` helper from `parseRule` to eliminate duplicated GA value
  validation logic (used identically for both `GA=` and `GA!=` filter tokens)
- Reduce cyclomatic complexity of `parseRule` from 24 to 20

**Related issue(s)**

PR #3302, opened 2026-06-12
Simplify matchState by Returning Conditions Directly

jaroslaw-pieszka:
**Description**

- Simplify `(*instances).matchState` in the in-memory storage driver by replacing
  `if <condition> { return true }` patterns with direct `return <condition>` in each
  switch case
- Reduce cyclomatic complexity of `matchState` from 30 to 21

**Related issue(s)**


piotrek1459 (approved):
Watch out for conflicts. Probably first #3301 then rebase at this and drop changes from blocklist.go

PR #3337, opened 2026-06-23
Available zones discovery for azure

piotrek1459:
**Description**

Extends KEB zone discovery to support Azure in addition to AWS.

New internal/hyperscalers/azure/ package:

AzureClient — per-call client using Azure ResourceSKUs.List API; filters zones by removing restrictions[type=Zone]; retries 3×2s on failure; local cache per client instance to deduplicate machine type queries within one operation
AzureCache — global background cache filled lazily at startup for all configured regions; refreshes every hour; re-fetches Gardener secret on each refresh to handle credential rotation; sync.RWMutex for concurrent access
AzureCachedClient — zero-latency client backed by global cache; used for HTTP validation and machines_availability
Factory dispatch (internal/hyperscalers/factory.go): Azure path returns AzureCachedClient if cache is ready for region, falls back to AzureClient otherwise. Async provisioning step always uses AzureClient with the Kyma-specific secret for accuracy.

machines_availability endpoint now dynamically iterates providers with zonesDiscovery: true instead of hardcoded AWS-only list.

MachineFamily for Azure (provider.go): extracts family prefix from Standard_<Family><size> pattern — e.g. Standard_D4s_v5 → D, Standard_NC4as_T4_v3 → NC.

Logging — unified format for both AWS and Azure: discovering/validating zones using credentials binding <name> region=<region>.


cmd/broker/main.go
```diff
@@ -379,9 +382,14 @@ func main() {
...
 	workersProvider := workers.NewProvider(cfg.InfrastructureManager, providerSpec, cfg.Broker.WorkerPoolLabelsAnnotationsEnabled)
 
-	factory := hyperscalers.NewFactory(providerSpec)
+	// azureSecretFetcher re-fetches the Azure secret on every cache refresh
+	// to handle credential rotation without restarting KEB.
+	azureSecretFetcher, err := buildAzureSecretFetcher(gardenerClient, rulesService)
+	if err != nil {
+		log.Warn(fmt.Sprintf("Azure zone cache unavailable, falling back to per-call mode: %s", err))
+	}
+	factory := hyperscalers.NewFactoryWithAzureCache(ctx, providerSpec, azureSecretFetcher)
```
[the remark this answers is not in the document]
piotrek1459: Fixed — gated the entire block on `providerSpec.ZonesDiscovery(pkg.Azure)` so the warning is only emitted when Azure discovery is actually configured. Used a separate local error variable to avoid polluting the outer `err`.

internal/hyperscalers/azure/cache.go
```diff
@@ -0,0 +1,190 @@
...
+				break
+			} else {
+				lastErr = err
+				time.Sleep(cacheRetryInterval)
+			}
+		}
+		if lastErr != nil {
+			slog.Error(fmt.Sprintf("failed to fill Azure zone cache for region %s after %d retries: %s", region, cacheRetries, lastErr))
+		}
+	}
```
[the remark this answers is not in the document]
piotrek1459: Fixed — replaced the `lastErr != nil` check with a `filled` boolean flag so the error is only logged when all retries are exhausted, not when a retry succeeds.

internal/hyperscalers/azure/cache.go
```diff
@@ -0,0 +1,190 @@
...
+
+	regions := c.providerSpec.Regions(pkg.Azure)
+	for _, region := range regions {
+		var lastErr error
+		for i := 0; i < cacheRetries; i++ {
+			if err := c.fillRegion(ctx, secret, region); err == nil {
+				break
+			} else {
+				lastErr = err
+				time.Sleep(cacheRetryInterval)
```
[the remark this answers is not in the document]
piotrek1459: Fixed — replaced `time.Sleep(cacheRetryInterval)` with a `select { case <-time.After(cacheRetryInterval): case <-ctx.Done(): return }` so the loop respects context cancellation during retry backoff.

internal/hyperscalers/azure/client.go
```diff
@@ -0,0 +1,222 @@
...
+type ResourceSKUsAPI interface {
+	NewListPager(options *armcompute.ResourceSKUsClientListOptions) *runtime.Pager[armcompute.ResourceSKUsClientListResponse]
+}
+
+type AzureClient struct {
+	skusClient   ResourceSKUsAPI
+	region       string
+	providerSpec *configuration.ProviderSpec
+	cache        map[string][]string
+	cacheDone    bool
```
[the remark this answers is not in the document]
piotrek1459: `AzureClient` is not used concurrently — each instance is created per provisioning/update operation and `AvailableZones` is called sequentially within a single goroutine (the step's `Run` method iterates over machine types one at a time). No mutex needed. `go test -race` passes.

internal/hyperscalers/azure/client.go
```diff
@@ -0,0 +1,222 @@
...
+	for z := range available {
+		zones = append(zones, z)
+	}
+	return zones
+}
+
+func ExtractSubscriptionID(secret *unstructured.Unstructured) (string, error) {
+	data, found, err := unstructured.NestedStringMap(secret.Object, "data")
+	if err != nil || !found {
+		return "", fmt.Errorf("unable to extract data from secret: %w", err)
```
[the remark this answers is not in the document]
piotrek1459: Fixed — split into two separate conditions matching the pattern in `ExtractCredentials`: `err != nil` returns a wrapped error, `!found` returns `"secret does not contain data"`.

internal/hyperscalers/azure/cache_test.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+func buildCacheSpec(machineNames []string) *configuration.ProviderSpec {
+	machines := ""
+	for _, name := range machineNames {
+		machines += "    \"" + name + "\": \"" + name + "\"\n"
+	}
+	yaml := "azure:\n  zonesDiscovery: true\n  regions:\n    westeurope:\n      displayName: \"West Europe\"\n  machines:\n" + machines
+	spec, _ := configuration.NewProviderSpec(strings.NewReader(yaml))
+	return spec
```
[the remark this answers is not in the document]
piotrek1459: Fixed — added `panic` if `NewProviderSpec` returns an error so test failures are explicit rather than silent.

internal/provider/configuration/provider.go
```diff
@@ -186,10 +186,24 @@ func (p *ProviderSpec) ZonesDiscovery(cp runtime.CloudProvider) bool {
...
+	var providers []runtime.CloudProvider
+	for name, dto := range p.data {
+		if dto.ZonesDiscovery {
+			// Normalize to match runtime.CloudProvider constants (e.g., "aws" → "AWS")
+			normalized := runtime.CloudProviderFromString(string(name))
+			providers = append(providers, normalized)
+		}
+	}
+	return providers
+}
```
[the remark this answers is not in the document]
piotrek1459: Fixed — added `sort.Slice` to return providers in stable alphabetical order.

internal/hyperscalers/factory.go
```diff
@@ -6,22 +6,47 @@ import (
...
 type hyperscalerFactory struct {
 	providerSpec *configuration.ProviderSpec
+	azureCache   *azure.AzureCache
 }
 
+// NewFactory creates a new Factory. If an azureSecret is provided and Azure has zonesDiscovery
+// enabled, a global background cache is started for all configured Azure regions.
 func NewFactory(providerSpec *configuration.ProviderSpec) Factory {
 	return &hyperscalerFactory{providerSpec: providerSpec}
 }
```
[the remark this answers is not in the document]
piotrek1459: Fixed — updated the comment to accurately describe what `NewFactory` does (no cache, direct API calls).

internal/hyperscalers/factory.go
```diff
@@ -6,22 +6,47 @@ import (
...
+
 func (f *hyperscalerFactory) NewFromSecret(ctx context.Context, provider pkg.CloudProvider, secret *unstructured.Unstructured, region string) (ProviderClient, error) {
 	switch provider {
 	case pkg.AWS:
 		return aws.NewClientFromSecret(ctx, f.providerSpec, secret, region)
+	case pkg.Azure:
+		// Use global cache if available and ready for this region — zero latency.
+		// Falls back to per-call client if cache is not yet ready (lazy fill in progress).
+		if f.azureCache != nil && f.azureCache.Ready(region) {
+			return azure.NewCachedClient(f.azureCache, region, f.providerSpec), nil
```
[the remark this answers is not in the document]
piotrek1459: Acknowledged and documented. The cached client intentionally ignores the caller-provided secret — this is the explicit trade-off: cache = speed for HTTP validation and `machines_availability`, per-call client with exact Kyma secret = accuracy for the async `DiscoverAvailableZonesCBStep`. Added a comment in `NewFromSecret` explaining this.

internal/hyperscalers/azure/client.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+	return decodeField(data, "subscriptionID")
+}
```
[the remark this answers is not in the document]
piotrek1459: Fixed — split into two separate conditions matching the pattern in `ExtractCredentials`: `err != nil` returns a wrapped error, `!found` returns `"secret does not contain data"`.

internal/hyperscalers/azure/client.go
```diff
@@ -0,0 +1,222 @@
...
+	for i := 0; i < retries; i++ {
+		if err := c.tryFillCache(ctx); err == nil {
+			return nil
+		} else {
+			lastErr = err
+			c.cache = nil
+			time.Sleep(interval)
+		}
+	}
+	return lastErr
```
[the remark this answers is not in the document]
piotrek1459: `AzureClient` is short-lived (one per operation, seconds) and `fillCache` is only called from a single goroutine. The cost of propagating `ctx` through the retry loop outweighs the benefit for such a short-lived object. Left as `time.Sleep`. The `AzureCache` background goroutine (which is long-lived) has been fixed to use ctx-aware sleep.

internal/hyperscalers/azure/cache.go
```diff
@@ -0,0 +1,190 @@
...
+			}
+		}
```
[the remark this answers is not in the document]
piotrek1459: Fixed — replaced `time.Sleep` with a context-aware `select` so the retry backoff respects `ctx.Done()` and shutdown is not delayed.

cmd/broker/main.go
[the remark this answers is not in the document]
piotrek1459: Fixed — gated the `buildAzureSecretFetcher` call behind `providerSpec.ZonesDiscovery(pkg.Azure)` so it is only invoked when Azure discovery is actually enabled.

internal/broker/instance_create.go
```diff
@@ -1213,6 +1213,8 @@ func newHyperscalerClient(
 		return nil, fmt.Errorf("unable to get secret %s/%s: %w", credentialsBinding.GetSecretRefNamespace(), credentialsBinding.GetSecretRefName(), err)
 	}
 
+	log.Info(fmt.Sprintf("validating zones using credentials binding %s/%s region=%s", credentialsBinding.GetSecretRefNamespace(), credentialsBinding.GetSecretRefName(), values.Region))
```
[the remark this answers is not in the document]
piotrek1459: Fixed — changed the message from `"validating zones using credentials binding"` to `"validating zones using secret"` since what is logged is the secret ref (`namespace/name`), not the CredentialsBinding name.

cmd/broker/main.go
```diff
@@ -664,3 +677,34 @@ func resolvedMachineTypesForKCR(providerSpec *configuration.ProviderSpec, provid
...
+		Plan:        "azure",
+		Hyperscaler: "azure",
+	}
+	matchedRule, found := rulesService.MatchProvisioningAttributesWithValidRuleset(attr)
+	if !found {
+		return nil, fmt.Errorf("no matching rule for azure hyperscaler")
+	}
+	labelSelector := subscriptions.NewLabelSelectorFromRuleset(matchedRule).BuildAnySubscription()
+
+	return func() (*unstructured.Unstructured, error) {
```
jaroslaw-pieszka: Is this the only option? Returning *unstructured.Unstructured - we always return secret structure, not any structure.

internal/hyperscalers/azure/cache.go
```diff
@@ -0,0 +1,199 @@
...
+	refreshInterval    = 1 * time.Hour
+	cacheRetries       = 3
+	cacheRetryInterval = 2 * time.Second
+)
+
+// SecretFetcher fetches the current Azure credentials secret from Gardener.
+// Called on every cache refresh to pick up rotated credentials.
+type SecretFetcher func() (*unstructured.Unstructured, error)
+
+// SKUsClientFactory creates a ResourceSKUsAPI from credentials.
```
jaroslaw-pieszka: SKU - Stock-Keeping Unit - would it be easier to read with such explanation in the comments?

internal/hyperscalers/azure/cache.go
```diff
@@ -0,0 +1,199 @@
...
+	defer c.mu.RUnlock()
+	regionCache, ok := c.data[region]
+	if !ok {
+		return nil
+	}
+	return regionCache[machineType]
+}
+
+// Ready reports whether the cache has been filled for the given region.
+func (c *AzureCache) Ready(region string) bool {
```
jaroslaw-pieszka, line 83: It means there is at least one entry for region - so it could be only partially filled.

internal/hyperscalers/azure/cache.go
```diff
@@ -0,0 +1,199 @@
...
+	return armcompute.NewResourceSKUsClient(subscriptionID, credential, nil)
+}
+
+// AzureCache is a global cache of available zones per region and machine type.
+// It is filled lazily in the background after KEB starts and refreshed every hour.
+// Only regions and machine types configured in providerSpec are cached.
+// The secret is re-fetched on every refresh to handle credential rotation.
+type AzureCache struct {
+	mu                sync.RWMutex
+	data              map[string]map[string][]string // region → machineType → zones
```
jaroslaw-pieszka, line 49: We could consider flattening the structure and use compound key (region+machine type) unless there are scenarios when we use this map differently.

internal/hyperscalers/azure/cache.go
```diff
@@ -0,0 +1,199 @@
...
+			} else {
+				lastErr = err
+				select {
+				case <-time.After(cacheRetryInterval):
+				case <-ctx.Done():
+					return
+				}
+			}
+		}
+		if !filled {
```
jaroslaw-pieszka: filled means err=nil

internal/hyperscalers/azure/cache.go
```diff
@@ -0,0 +1,199 @@
...
+	supported := make(map[string]struct{})
+	for _, mt := range c.providerSpec.MachineTypes(pkg.Azure) {
+		supported[mt] = struct{}{}
+	}
+
+	regionCache := make(map[string][]string)
+	filter := fmt.Sprintf("location eq '%s'", region)
+	pager := skusClient.NewListPager(&armcompute.ResourceSKUsClientListOptions{Filter: &filter})
+
+	for pager.More() {
```
jaroslaw-pieszka, line 170: I would consider extracting this loop as separate function.

internal/hyperscalers/azure/cache.go
```diff
@@ -0,0 +1,199 @@
...
+	if err != nil {
+		return fmt.Errorf("while creating Azure credential: %w", err)
+	}
+
+	skusClient, err := c.skusClientFactory(subscriptionID, credential)
+	if err != nil {
+		return fmt.Errorf("while creating Azure ResourceSKUs client: %w", err)
+	}
+
+	supported := make(map[string]struct{})
```
jaroslaw-pieszka: supportedMachineTypes?

internal/hyperscalers/azure/client.go
```diff
@@ -0,0 +1,225 @@
...
+
+	available := make(map[string]struct{})
+	for _, z := range sku.LocationInfo[0].Zones {
+		if z != nil {
+			available[*z] = struct{}{}
+		}
+	}
+
+	for _, restriction := range sku.Restrictions {
+		if restriction.Type == nil || *restriction.Type != armcompute.ResourceSKURestrictionsTypeZone {
```
jaroslaw-pieszka, line 123: Is this guaranteed that only one restriction would have type `armcompute.ResourceSKURestrictionsTypeZone` ?

internal/hyperscalers/azure/client.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+
+	slog.Info(fmt.Sprintf("querying Azure ResourceSKUs for region %s", c.region))
+
+	filter := fmt.Sprintf("location eq '%s'", c.region)
+	pager := c.skusClient.NewListPager(&armcompute.ResourceSKUsClientListOptions{
+		Filter: &filter,
+	})
+
+	for pager.More() {
```
jaroslaw-pieszka: I have seen this code before. Is this a duplicate?

internal/hyperscalers/azure/cache.go
```diff
@@ -0,0 +1,199 @@
...
+		slog.Error(fmt.Sprintf("failed to fetch Azure secret for cache refresh: %s", err))
+		return
+	}
+
+	regions := c.providerSpec.Regions(pkg.Azure)
+	for _, region := range regions {
+		var filled bool
+		var lastErr error
+		for i := 0; i < cacheRetries; i++ {
+			if err := c.fillRegion(ctx, secret, region); err == nil {
```
jaroslaw-pieszka: If we encounter error with credentials extraction we will retry - this would be futile.

internal/hyperscalers/azure/cache_test.go
```diff
@@ -0,0 +1,226 @@
...
+		"data": map[string]interface{}{
+			"clientID":       enc("test-client-id"),
+			"clientSecret":   enc("test-client-secret"),
+			"tenantID":       enc("test-tenant-id"),
+			"subscriptionID": enc("test-subscription-id"),
+		},
+	}
+	return s
+}
+
```
jaroslaw-pieszka, line 26: Is there a case with fail during:
- credentials extraction
- azure credential creation?

internal/hyperscalers/azure/client.go
```diff
@@ -0,0 +1,225 @@
...
+	skusClient   ResourceSKUsAPI
+	region       string
+	providerSpec *configuration.ProviderSpec
+	cache        map[string][]string
+	cacheDone    bool
+}
+
+func NewClientFromSecret(ctx context.Context, providerSpec *configuration.ProviderSpec, secret *unstructured.Unstructured, region string) (*AzureClient, error) {
+	clientID, clientSecret, tenantID, subscriptionID, err := ExtractCredentials(secret)
+	if err != nil {
```
jaroslaw-pieszka, line 42: I have seen this code - code duplication.

internal/hyperscalers/azure/client.go
```diff
@@ -0,0 +1,225 @@
...
+		skusClient:   skusClient,
+		region:       region,
+		providerSpec: providerSpec,
+	}, nil
+}
+
+func (c *AzureClient) AvailableZones(ctx context.Context, machineType string) ([]string, error) {
+	machineType = c.providerSpec.ResolveMachineType(pkg.Azure, machineType)
+
+	if err := c.fillCache(ctx); err != nil {
```
jaroslaw-pieszka: This is slightly cryptic/obscure and naming does not help.

We call fillCache, but we do not fill cache if it is `done`. And this `done` is vague. Caching is done or more precisely cache is up-to-date pre-filled, filled, but cache is not done.

internal/hyperscalers/azure/client.go
```diff
@@ -0,0 +1,225 @@
...
+			}
+			if _, ok := supported[*sku.Name]; !ok {
+				continue
+			}
+
+			c.cache[*sku.Name] = availableZonesFromSKU(sku)
+		}
+	}
+
+	c.cacheDone = true
```
jaroslaw-pieszka: Even if there is not any Page, we will set cacheDone (cacheFilled) to true.

internal/hyperscalers/azure/client.go
```diff
@@ -0,0 +1,225 @@
...
+	if err != nil {
+		return "", fmt.Errorf("unable to extract data from secret: %w", err)
+	}
+	if !found {
+		return "", fmt.Errorf("secret does not contain data")
+	}
+	return decodeField(data, "subscriptionID")
+}
+
+func ExtractCredentials(secret *unstructured.Unstructured) (clientID, clientSecret, tenantID, subscriptionID string, err error) {
```
jaroslaw-pieszka: Wouldn't structure for Credential be handy?

internal/hyperscalers/azure/cache.go
```diff
@@ -0,0 +1,199 @@
...
+func defaultSKUsClientFactory(subscriptionID string, credential *azidentity.ClientSecretCredential) (ResourceSKUsAPI, error) {
+	return armcompute.NewResourceSKUsClient(subscriptionID, credential, nil)
+}
+
+// AzureCache is a global cache of available zones per region and machine type.
+// It is filled lazily in the background after KEB starts and refreshed every hour.
+// Only regions and machine types configured in providerSpec are cached.
+// The secret is re-fetched on every refresh to handle credential rotation.
+type AzureCache struct {
+	mu                sync.RWMutex
```
jaroslaw-pieszka, line 48: https://jisho.org/search/無%20%23kanji - disturbing

cmd/broker/main.go
```diff
@@ -664,3 +677,34 @@ func resolvedMachineTypesForKCR(providerSpec *configuration.ProviderSpec, provid
 	}
 	return result
 }
+
+// buildAzureSecretFetcher returns a SecretFetcher that fetches the first available Azure
+// credentials secret from Gardener on each call. This ensures credential rotation is
+// picked up on every cache refresh without restarting KEB.
+func buildAzureSecretFetcher(gardenerClient *gardener.Client, rulesService *rules.RulesService) (azurehyperscaler.SecretFetcher, error) {
```
jaroslaw-pieszka, line 690: I suspect almost whole body of this function is copied from ResolveCB step, but here we provide hard-wired ProvisioningAttributess. So could we extract common part to avoid duplication?

internal/broker/instance_create.go
```diff
@@ -1218,5 +1218,12 @@ func newHyperscalerClient(
 		return nil, fmt.Errorf("unable to create hyperscaler client: %w", err)
 	}
 
+	type cacheClient interface{ IsFromCache() bool }
```
jaroslaw-pieszka: Could you explain why this function is necessary? Should we be able to distinguish if we are using cache or not?

internal/hyperscalers/azure/client.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+
+	c.cacheDone = true
+	slog.Info(fmt.Sprintf("Azure ResourceSKUs loaded for region %s (%d machine types cached)", c.region, len(c.cache)))
+	return nil
+}
+
+// availableZonesFromSKU returns zones where the SKU is available,
+// excluding zones with zone-type restrictions.
+func availableZonesFromSKU(sku *armcompute.ResourceSKU) []string {
```
jaroslaw-pieszka, line 116: If AZones are decimal numbers (represented as strings) this approach is overkill.

internal/hyperscalers/azure/client.go
```diff
@@ -0,0 +1,225 @@
...
+	if len(sku.LocationInfo) == 0 {
+		return nil
+	}
+
+	available := make(map[string]struct{})
+	for _, z := range sku.LocationInfo[0].Zones {
```
jaroslaw-pieszka: Comment could be helpful - why we use hard-wired magic number 0 as index?

internal/hyperscalers/azure/client.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+	subscriptionID, err = decodeField(data, "subscriptionID")
+	if err != nil {
+		return "", "", "", "", err
+	}
+
+	return clientID, clientSecret, tenantID, subscriptionID, nil
+}
+
+func decodeField(data map[string]string, field string) (string, error) {
```
jaroslaw-pieszka: This function not only decodes but checks presence as well. Consider separation of responsibilities, two functions or even more.
Separating validation (presence of required fields) from decoding.

internal/hyperscalers/azure/cache.go
piotrek1459: Will think of situation of not ready region. Maybe len(map) is required.

cmd/broker/main.go
piotrek1459: Think of better way (remove duplication), maybe extract it somewhere.

internal/broker/instance_create.go
piotrek1459: Maybe we can avoid such expressions.

internal/hyperscalers/azure/cache.go
piotrek1459: After discussion we decided to leave it.

cmd/broker/main.go
piotrek1459: Good point. Changed SecretFetcher to return a new AzureCredentials struct instead of *unstructured.Unstructured. The internal GetSecret helper still returns *unstructured.Unstructured (changing it globally would touch many unrelated callsites), so buildAzureSecretFetcher in main.go now calls ExtractCredentials and returns the decoded struct. This way the cache only ever sees typed credentials, not raw secret data.

internal/hyperscalers/azure/cache.go
piotrek1459: Good idea. Added the expansion to the comment: SKUsClientFactory creates a ResourceSKUsAPI (SKU = Stock-Keeping Unit) from credentials.

internal/hyperscalers/azure/cache.go
piotrek1459: Ready(region) returns true only after a successful fillRegion call — if all retries fail, the region is never written to c.data and Ready returns false. In that case NewFromSecret in the factory falls back to a per-call NewClientFromSecret (live API). So a partially-filled cache is not a problem: each region is independently either fully cached or fully live. The cache retries once per hour, so a failed region gets another chance on the next refresh cycle.

The comment on Ready was misleading — updated it to reflect this.

internal/hyperscalers/azure/cache.go
piotrek1459: Extracted the pager loop into fetchZonesBySKU(ctx, skusClient, region, supportedMachineTypes) — fillRegion now only handles credential setup, client creation, and cache write

internal/hyperscalers/azure/cache.go
piotrek1459: supported → supportedMachineTypes

internal/hyperscalers/azure/cache.go
piotrek1459: Extracted retry loop from fillAll into fillRegionWithRetry — fillAll is now flat, filled bool is gone

internal/hyperscalers/azure/client.go
piotrek1459: Not guaranteed by the Azure API spec — a SKU can have multiple restrictions of type Zone, each with a different reasonCode (e.g. NotAvailableForSubscription vs NotAvailableForRequest) and a different set of zones. The loop iterates over all of them regardless of reason code, so all restricted zones are removed from the available set. Added a comment to availableZonesFromSKU to make this explicit.

internal/hyperscalers/azure/client.go
piotrek1459: Yes, it was a duplicate — tryFillCache had its own inline pager loop with identical filtering logic. Extracted to fetchZonesBySKU in cache.go (same package), which is now the single place for SKU listing and zone extraction. tryFillCache now delegates to it.

internal/hyperscalers/azure/cache.go
piotrek1459: Fixed. Credential creation (NewClientSecretCredential + skusClientFactory) is now done once in fillRegionWithRetry before the retry loop — if credentials are invalid, the function returns immediately without retrying. Only the actual API call (fetchZonesBySKU) is retried.

This also improved the tests — fillRegion now takes a ResourceSKUsAPI directly instead of AzureCredentials, so tests pass a mock API without going through any credential creation. Previously tests were calling azidentity.NewClientSecretCredential with fake values, which worked only because the SDK doesn't validate credentials until the first HTTP call.

internal/hyperscalers/azure/cache_test.go
piotrek1459: Added TestAzureCache_SecretFetcherError (credentials extraction failure — already existed) and TestAzureCache_ClientFactoryError_NoRetry (client creation failure — asserts that skusClientFactory is called exactly once, i.e. not retried). NewClientSecretCredential itself can't fail with non-empty inputs, so the factory error is the closest observable signal for the credential creation path.

internal/hyperscalers/azure/client.go
piotrek1459: ExtractCredentials is called in both NewClientFromSecret and buildAzureSecretFetcher, but the contexts are different — NewClientFromSecret creates a ready-to-use client immediately, while buildAzureSecretFetcher returns a closure that fetches and decodes credentials on each cache refresh. ExtractCredentials is a shared helper for decoding *unstructured.Unstructured secrets, so it's expected to appear wherever credentials are decoded — imo there's no logic to extract.

internal/hyperscalers/azure/client.go
piotrek1459: Renamed: cacheDone → cacheLoaded, fillCache → ensureZonesLoaded. The new names make the intent clear — cacheLoaded is a state flag, and ensureZonesLoaded is an idempotent operation that loads zones if not already done.

internal/hyperscalers/azure/client.go
piotrek1459: This is now handled in fetchZonesBySKU — if the pager returns no pages, it returns an empty map and nil error. cacheLoaded is set to true after that, which is correct: it means the API was queried and the result (even if empty) is cached. An empty result for a region means no configured machine types are available there — that's a valid cache state, not a failure.

internal/hyperscalers/azure/client.go
piotrek1459: Good idea — ExtractCredentials now returns (AzureCredentials, error) instead of 5 separate values. All callers updated: NewClientFromSecret and buildAzureSecretFetcher in main.go are now simpler, and the struct was already there since the SecretFetcher change.

cmd/broker/main.go
piotrek1459: You're right that both share the same core pattern: MatchProvisioningAttributesWithValidRuleset → labelSelector → GetCredentialsBindings → GetSecret. However ResolveCredentialsBindingStep has significantly more logic on top (claiming, multi-account support, shared subscriptions, retry). The common part could be extracted as a helper on gardener.Client (e.g. GetFirstSecretForAttributes), but that would introduce a new public method for a single use case. Happy to do it if you think it's worth the abstraction — otherwise we can leave it as is since the two contexts are quite different.
@jaroslaw-pieszka

internal/broker/instance_create.go
piotrek1459: Removed IsFromCache() and the inline interface check. The cache and per-call client already log at the point of use, so the distinction was redundant. Replaced with a single log line showing region and secret ref.

internal/hyperscalers/azure/client.go
piotrek1459: Good point. Simplified to a single restricted set — first collect all restricted zones, then filter LocationInfo[0].Zones keeping only those not in the set. One map instead of two, original API order preserved, and easier to read. Order of returned zones doesn't matter for us — zones are shuffled later when building the Runtime CR worker config.

internal/hyperscalers/azure/client.go
piotrek1459: Added a comment. LocationInfo[0] is safe because the pager is filtered by location eq '<region>', so the Azure API always returns exactly one LocationInfo entry — the one for the queried region.

internal/hyperscalers/azure/client.go
piotrek1459: Renamed decodeField → extractField — the new name describes what it actually does: extracts a field from the data map (checks presence + decodes base64). Splitting into two separate functions would add complexity without real benefit since the callers always need both steps together.

internal/hyperscalers/azure/client.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+	return len(zones), nil
+}
+
+func (c *AzureClient) ensureZonesLoaded(ctx context.Context) error {
+	if c.cacheLoaded {
+		return nil
+	}
+
+	var lastErr error
```
jaroslaw-pieszka: `lastErr` is superfluous.

internal/provider/configuration/provider.go
```diff
@@ -155,7 +155,7 @@ func (p *ProviderSpec) MachineDisplayNames(cp runtime.CloudProvider, machines []
 func (p *ProviderSpec) ValidateZonesDiscovery() error {
 	for provider, providerDTO := range p.data {
 		if providerDTO.ZonesDiscovery {
-			if provider != "aws" {
+			if provider != "aws" && provider != "azure" {
```
jaroslaw-pieszka:
```
  common/runtime.Azure             // "Azure"                               
  common/runtime.AWS               // "AWS"                                                                                                                                                                      
  common/runtime.GCP               // "GCP"                                                                                                                                                                      
  common/runtime.SapConvergedCloud // "SapConvergedCloud"                                                                                                                                                        
  common/runtime.Alicloud          // "Alicloud"                                                                                                                                                                 
  common/runtime.UnknownProvider   // "unknown"                                                                                                                                                                  
                                                          
```

internal/hyperscalers/azure/client.go
piotrek1459: Fixed — lastErr replaced with a single err variable declared before the loop and assigned directly in the condition.

internal/provider/configuration/provider.go
piotrek1459: The YAML keys ("aws", "azure") are lowercase as defined in providersConfig.yaml, while runtime.AWS/runtime.Azure constants are uppercase ("AWS", "Azure"). Since runtime.CloudProvider is just type CloudProvider string without custom UnmarshalYAML, the YAML decoder inserts keys as-is without normalization. Using CloudProviderFromString(string(provider)) handles the case mismatch. A cleaner fix would be adding UnmarshalYAML to CloudProvider to normalize at parse time, but that would affect all YAML parsing across the codebase — left for a separate change.

jaroslaw-pieszka (dismissed):

docs/contributor/03-55-zones-discovery.md
```diff
@@ -7,15 +7,16 @@
 The Zones Discovery feature extends Kyma Environment Broker (KEB) to dynamically determine availability zones for both the Kyma worker node pool and additional worker node pools during provisioning and updates.
 Operators can configure worker node pools to use either static zone assignments (predefined in the configuration) or dynamic zone assignments (queried live from the hyperscaler).
 
-> ### Note:
-> This feature is currently supported only on AWS.
+This feature is supported on **AWS** and **Azure**.
```
IwonaLanger:
```suggestion
> ### Note:
> This feature is only supported on Amazon Web Services and Microsoft Azure.
```

docs/contributor/03-55-zones-discovery.md
```diff
@@ -24,9 +25,13 @@ Example log entries:
 
 ```json lines
 {"level":"WARN", "msg":"Provider aws has zones discovery enabled, but region us-west-2 is configured with 4 static zone(s), which will be ignored."} 
-{"level":"WARN", "msg":"Provider aws has zones discovery enabled, but machine type g6 in region ap-south-1 is configured with 1 static zone(s), which will be ignored."}
+{"level":"WARN", "msg":"Provider azure has zones discovery enabled, but region westeurope is configured with 3 static zone(s), which will be ignored."}
 ```
 
+## Credentials
+
+KEB resolves hyperscaler credentials from Gardener secrets referenced by a `CredentialsBinding`. The secret fields follow the Gardener convention for each provider type.
```
IwonaLanger:
```suggestion
KEB resolves hyperscaler credentials from Gardener Secrets referenced by a `CredentialsBinding`. The Secret fields follow the Gardener convention for each provider type.
```

IwonaLanger: How about rephrasing to "The fields within a Secret" for more clarity? If that's what you mean.

docs/contributor/03-55-zones-discovery.md
```diff
@@ -36,6 +41,10 @@ During provisioning and updates, KEB validates the worker node pool configuratio
 To optimize performance, if the same machine type is used in multiple worker node pools, KEB queries the hyperscaler only once per unique machine type and reuses the result across all occurrences. This solution eliminates unnecessary duplicate calls.
 The subscription secret is used only for validation. Its name is logged to support traceability in case of validation failures.
 
+### Azure — zone restrictions
```
IwonaLanger:
```suggestion
### Microsoft Azure — Zone Restrictions
```

docs/contributor/03-55-zones-discovery.md
IwonaLanger: Let's stick to the Note format for better visibility.

docs/contributor/03-55-zones-discovery.md
piotrek1459: sure

notable-changes-to-release/notable-change.md
```diff
@@ -0,0 +1,41 @@
+<!--{"metadata":{"requirement":"RECOMMENDED","type":"INTERNAL","category":"CONFIGURATION"}}-->
+
+# KEB: Azure Availability Zone Discovery
+
+> ### Note:
+> No action is required to keep existing behavior. Zone discovery for Azure is opt-in and disabled by default. Enable it only if you want KEB to determine available zones dynamically from the Azure ResourceSKUs API instead of using static zone assignments.
```
IwonaLanger:
```suggestion
> No action is required to keep the existing behavior. Zone discovery for Microsoft Azure is opt-in and disabled by default. Enable it only if you want KEB to determine available zones dynamically from the Azure ResourceSKUs API instead of using static zone assignments.
```

notable-changes-to-release/notable-change.md
```diff
@@ -0,0 +1,41 @@
...
+
+## What's Changed
+
+KEB now supports dynamic availability zone discovery for Microsoft Azure, consistent with the existing AWS behavior.
+
+When enabled, KEB queries the Azure ResourceSKUs API at provisioning and update time to determine which availability zones are actually available for the requested machine type and subscription. Zones with Azure-level restrictions (`restrictions[type=Zone]`) are automatically excluded.
```
IwonaLanger:
```suggestion
When enabled, KEB queries the Azure ResourceSKUs API at provisioning and update time to determine which availability zones are available for the requested machine type and subscription. Zones with Azure-level restrictions (`restrictions[type=Zone]`) are automatically excluded.
```

notable-changes-to-release/notable-change.md
```diff
@@ -0,0 +1,41 @@
...
+
+Key behaviors:
+
+- The Kyma worker node pool requires at least 3 available zones. Provisioning is rejected synchronously if the machine type does not meet this requirement in the requested region.
+- Additional worker node pools require 3 zones for high-availability pools, or at least 1 zone for non-HA pools.
```
IwonaLanger:
```suggestion
- Additional worker node pools require 3 zones for high-availability (HA) pools, or at least 1 zone for non-HA pools.
```

notable-changes-to-release/notable-change.md
```diff
@@ -0,0 +1,41 @@
...
+- If the same machine type appears in multiple worker pools, the API is called only once per unique machine type.
+- A global background cache is populated at KEB startup and refreshed every hour, eliminating per-request latency for HTTP validation.
+- If both static zone configuration and `zonesDiscovery: true` are provided, a warning is logged and static zones are ignored.
+
+## Procedure
+
+To enable Azure zone discovery, set `zonesDiscovery: true` in the `providersConfig.yaml` ConfigMap under the `azure` section:
```
IwonaLanger:
```suggestion
To enable Azure zone discovery, set `zonesDiscovery: true` in the `providersConfig` ConfigMap under the `azure` section:
```

notable-changes-to-release/notable-change.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Post-Update Steps
+
+Monitor KEB startup logs for cache fill confirmation:
+
+```json
+{"level":"INFO","msg":"Azure zone cache filled for region westeurope (24 machine types)"}
+```
+
+If credentials are unavailable at startup, KEB falls back to per-call mode and logs a warning. Zone discovery still works but without the latency benefit of the global cache. Cache refreshes in 1 hour interval.
```
IwonaLanger:
```suggestion
If credentials are unavailable at startup, KEB falls back to per-call mode and logs a warning. Zone discovery still works, but without the latency benefit of the global cache. Cache refreshes at one-hour intervals.
```

IwonaLanger: Or did you mean "in one hour"?

jaroslaw-pieszka (approved):

IwonaLanger (approved):

PR #3365, opened 2026-07-06
Log SHA256 Digest of System CA Bundle at Startup

jaroslaw-pieszka:
**Description**

- Add `logCACertBundleDigestFromPaths` helper that iterates the cert file paths searched by Go's `crypto/x509` on Linux, finds the first non-empty file, and logs its path and SHA256 digest
- Log a warning with per-path status (missing/empty/error) if no usable CA bundle is found
- Call the helper at the end of `logConfiguration()` so it appears alongside other startup config logs

**Related issue(s)**

### DIFF
--- a/cmd/broker/main.go
+++ b/cmd/broker/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"crypto/fips140"
+	"crypto/sha256"
 	"fmt"
 	"log/slog"
 	"net/http"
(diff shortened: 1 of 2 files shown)


cmd/broker/main.go
```diff
@@ -504,6 +505,55 @@ func logConfiguration(logs *slog.Logger, cfg Config) {
...
+		logs.Info("CA bundle digest", slog.String("path", p), slog.String("sha256", fmt.Sprintf("%x", sum)))
+		return
+	}
+
+	// No usable bundle found — enumerate all paths with their status.
+	args := []any{slog.String("msg_detail", "no usable system cert bundle found")}
+	for _, s := range statuses {
+		args = append(args, slog.String(s.path, s.status))
+	}
+	logs.Warn("CA bundle digest", args...)
```
piotrek1459, line 556: Using the filesystem path as a structured-log key is an anti-pattern — it creates a new field per path per startup, which is noisy for log shippers.

Consider logging one line per path instead of building a slice:

```
for _, s := range statuses {
    logs.Warn("CA bundle digest",
        slog.String("path", s.path),
        slog.String("status", s.status),
    )
}
```
This also avoids the args... pattern which log/slog docs flag as losing static type safety.

jaroslaw-pieszka: There are two distinct code paths:                                                                                                                                                                             
                                                                                                                                                                                                                 
  Happy path (line 547) — a usable bundle is found:                                                                                                                                                              
```
  logs.Info("CA bundle digest", slog.String("path", p), slog.String("sha256", ...))
```
  This is perfectly fine. Uses typed slog.Attr directly. The output you showed ("path":..., "sha256":...) comes from this line. No issue here.                                                                   
                                                                                                                                                                                                                 
  Error path (lines 552-556) — no usable bundle found:                                                                                                                                                           
```
  args := []any{slog.String("msg_detail", "no usable system cert bundle found")}                                                                                                                                 
  for _, s := range statuses {                                                                                                                                                                                   
      args = append(args, slog.String(s.path, s.status))  // ← path used as KEY                                                                                                                                  
  }                                                                              
  logs.Warn("CA bundle digest", args...)                                                                                                                                                                         
```
                                                                                 
  This is where the reviewer's concern is valid — s.path (e.g. /etc/ssl/certs/ca-certificates.crt) is used as the field name/key, not as a value. That means each startup produces different field names, which  
  breaks log shippers that expect a fixed schema.

jaroslaw-pieszka: Sample output:
```
{"time":"2026-07-06T18:43:34.073628342Z","level":"INFO","msg":"CA bundle digest","path":"/etc/ssl/certs/ca-certificates.crt","sha256":"b8d837841b88bfaa1a0fa827cbca8e2576418dd47c9fc4bb7f1f9d89c83111b9"}
```

piotrek1459 (approved):
/lgtm

Issue #3352, opened 2026-07-01
Handle Gen2 image versions for Azure


KsaweryZietara:
**Description**

While testing the new Azure machine types `Standard_Ds_v6`, we encountered the following error:

```
The selected VM size 'Standard_D2s_v6' cannot boot Hypervisor Generation '1'. If this was a Create operation please check that the Hypervisor Generation of the Image matches the Hypervisor Generation of the selected VM Size. If this was an Update operation please select a Hypervisor Generation '1' VM Size. For more information, see https://aka.ms/azuregen2vm
```

During Runtime creation, KEB sets the machine image version in the Runtime CR using the value from the Infrastructure Manager configuration:

```
	provider := imv1.Provider{
		Workers: []gardener.Worker{
			{
				Name: "cpu-worker-0",
				Machine: gardener.Machine{
					Type: s.providerSpec.ResolveMachineType(
						pkg.CloudProviderFromString(values.ProviderType),
						DefaultIfParamNotSet(values.DefaultMachineType, operation.ProvisioningParameters.Parameters.MachineType),
					),
					Image: &gardener.ShootMachineImage{
						Name:    s.config.MachineImage,
						Version: &s.config.MachineImageVersion,
					},
				},
			},
		},
	}
```

Current configuration:

```
  infrastructureManager:
    machineImage: gardenlinux
    machineImageVersion: 2150.3.0
```

Some Azure machine types require a Gen2-compatible image version, for example `2150.3.0-gen2`, instead of the default Gen1 image version. See [Support for Generation 2 VMs on Azure](https://learn.microsoft.com/en-us/azure/virtual-machines/generation-2)

KEB should determine the supported Azure image generations per Azure machine type by querying the Azure API. During provisioning or update operations, including new worker pool creation, KEB should select the newest supported generation.

The existing `machineImage` and `machineImageVersion` configuration must remain unchanged. Instead, KEB should derive the effective image version when required, for example: `2150.3.0-gen2`.

For details, see the refinement arrangements: https://github.com/kyma-project/kyma-environment-broker/issues/3352#issuecomment-4852913915

PK85:
some comments after planning:
- looks like we can use gen2 as default for all machineTypes, [r Generation 2 VMs on Azure](https://learn.microsoft.com/en-us/azure/virtual-machines/generation-2). We need to test on dev. Then create a ticket to SRE to migrate all azure instances to gen2 (Keb must alreadyt work with gen2 before migration)
[...]

PR #3396, opened 2026-07-15
Add Azure Machine Image Version Suffix Based on HyperV Generation

KsaweryZietara:
**Description**

- Add `HyperVGeneration` method to `ProviderClient` interface and implement it in `AzureClient`, returning a `-gen2` suffix when the machine type supports HyperV Generation V2
- Store per-machine-type image version suffixes on the operation during zone discovery (`DiscoverAvailableZonesCBStep`)
- Apply the suffix to the machine image version for the Kyma worker node pool during provisioning and update
- Apply the suffix to additional worker node pools via `workers.Provider`, preserving the existing version when a pool is unchanged
- Pass `machineImageVersionSuffix` flag through `NewFactoryWithAzureCache` and `AzureClient` so the feature is gated at the client level
- Add `MachineImageVersionSuffix` config field to `InfrastructureManager`, `deployment.yaml`, `values.yaml`, and generated docs

**Related issue(s)**
See also #3352


cmd/broker/broker_suite_test.go
```diff
@@ -212,7 +218,10 @@ func newBrokerSuiteTest(t *testing.T, o *suiteOptions) *BrokerSuiteTest {
 		require.Empty(t, rulesService.ValidationInfo.PlanErrors)
 	}
 
-	factory := fixture.NewFakeFactory(fixDiscoveredZones(), nil)
+	factory := hyperscalers.Factory(fixture.NewFakeFactory(fixDiscoveredZones(), nil))
```
jaroslaw-pieszka: So this factory is overwritten with SuiteOption.
If factory SuiteOption is present we can skip calling hyperscalers.Factory(), can't we?

internal/broker/broker.go
```diff
@@ -179,6 +179,7 @@ type InfrastructureManager struct {
 	ControlPlaneFailureTolerance string            `envconfig:"optional"`
 	UseSmallerMachineTypes       bool              `envconfig:"default=false"`
 	IngressFilteringPlans        StringList        `envconfig:"default=no-plan"`
+	MachineImageVersionSuffix    bool              `envconfig:"default=false"`
```
jaroslaw-pieszka: One reservations (but this is nit picking).
Variable name suggests that it contains suffix itself, not information if suffix is supposed to be added/used.

cmd/broker/broker_suite_test.go
KsaweryZietara: Applied the fix. The default factory is now only constructed when no `WithFactory` option is provided.

internal/broker/broker.go
KsaweryZietara: Do you think `MachineImageVersionSuffixEnabled` is a better name?

jaroslaw-pieszka: Or Use...Suffix. Do you think it makes sense?

KsaweryZietara: Yes, I’ll use the name `UseMachineImageVersionSuffix`.

resources/keb/values.yaml
```diff
@@ -394,6 +394,8 @@ infrastructureManager:
   maxPods: 200
   # If true, enables provisioning of clusters with nodes distributed across multiple availability zones.
   multiZoneCluster: "true"
+  # If true, appends a machine-type-specific suffix (e.g. "-gen2") to the machine image version for Azure workers.
```
IwonaLanger:
```suggestion
  # If true, appends a machine-type-specific suffix (for example, "-gen2") to the machine image version for Azure workers.
```

jaroslaw-pieszka (dismissed):

jaroslaw-pieszka (dismissed):

IwonaLanger (approved):

jaroslaw-pieszka (approved):

PR #3409, opened 2026-07-21
Migrate to github apps

piotrek1459:
**Description**

- As part of hardening we are migrating our workflows to use GitHub apps to authenticate instead of service account.
- POC is here https://github.com/kyma-project/kyma-environment-broker/actions/runs/29805239960
**Related issue(s)**


.github/workflows/create-release.yaml
```diff
@@ -468,9 +476,17 @@ jobs:
           git commit -m "Update KEB chart metadata with version ${{ inputs.name }}"
           git push origin gh-pages
 
+      - name: Generate bot token
```
jaroslaw-pieszka, line 479: Again?

.github/workflows/sync-docs-toc.yml
```diff
@@ -242,7 +253,7 @@ jobs:
           git commit -m "docs(keb): sync doc structure changes to toc.yaml
 
           Triggered by kyma-environment-broker@${GITHUB_SHA}"
-          git remote set-url origin ${PRODUCT_RUNTIME_CLONE_URL}
+          git remote set-url origin "https://x-access-token:${{ steps.enterprise-bot-token.outputs.token }}@${{ vars.GH_TOOLS_HOST }}/kyma/${{ vars.PK_RUNTIME_REPO_NAME }}.git"
```
jaroslaw-pieszka: Why do we replace variables with so complex expressions?

.github/workflows/create-release.yaml
piotrek1459: Step outputs are not accessible across jobs. The bot-token generated in the bumps job is not available in publish-release — these are separate runners. Passing tokens between jobs via job outputs would expose them in logs. Generating a fresh token per job is the recommended pattern for actions/create-github-app-token — tokens are short-lived (1h) and scoped anyway.

.github/workflows/sync-docs-toc.yml
piotrek1459: x-access-token is the standard username for GitHub App installation tokens in HTTPS clone URLs — GitHub expects this specific string. The token itself comes from actions/create-github-app-token and is short-lived (1h). I moved the full URL to a step-level env var (MANAGEMENT_PLANE_CHARTS_WITH_USER) to avoid repeating it inline.

jaroslaw-pieszka (approved):

PR #3418, opened 2026-07-22
Fetch Azure Machine Image Version Suffixes Independently of Zone Discovery

KsaweryZietara:
**Description**

- Decouple HyperV generation suffix fetching from zone discovery in `DiscoverAvailableZonesCBStep` — suffixes are now always fetched for Azure regardless of the `zonesDiscovery` flag
- Skip zone fetching per machine type when zones discovery is disabled, while still running the step for Azure
- Add `NewAzureProviderSpec` fixture helper to allow configuring `zonesDiscovery` flag in tests
- Add `sb-azure-eu-access` secret to `fixSecrets` in the broker suite test to support Azure EU-access provisioning flows that now reach the step
- Add test `TestDiscoverAvailableZonesCBStep_AzureZonesDiscoveryDisabledStillFetchesSuffixes`

**Related issue(s)**
See also #3352

### DIFF
--- a/internal/process/steps/discover_available_zones_cb.go
+++ b/internal/process/steps/discover_available_zones_cb.go
@@ -43,11 +43,13 @@ func (s *DiscoverAvailableZonesCBStep) Name() string {
 
 func (s *DiscoverAvailableZonesCBStep) Run(operation internal.Operation, log *slog.Logger) (internal.Operation, time.Duration, error) {
 	provider := runtime.CloudProviderFromString(operation.ProviderValues.ProviderType)
-	if !s.providerSpec.ZonesDiscovery(provider) {
+	zonesDiscoveryEnabled := s.providerSpec.ZonesDiscovery(provider)
+
+	if !zonesDiscoveryEnabled && provider != runtime.Azure {
 		log.Info(fmt.Sprintf("Zones discovery disabled for provider %s, skipping", provider))
 		return operation, 0, nil
 	}
-	if len(operation.DiscoveredZones) > 0 {
+	if zonesDiscoveryEnabled && len(operation.DiscoveredZones) > 0 {
 		log.Info("Available zones already discovered, skipping")
 		return operation, 0, nil
 	}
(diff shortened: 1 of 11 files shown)


internal/process/steps/discover_available_zones_cb.go
```diff
@@ -43,12 +45,17 @@ func (s *DiscoverAvailableZonesCBStep) Name() string {
 
 func (s *DiscoverAvailableZonesCBStep) Run(operation internal.Operation, log *slog.Logger) (internal.Operation, time.Duration, error) {
 	provider := runtime.CloudProviderFromString(operation.ProviderValues.ProviderType)
-	if !s.providerSpec.ZonesDiscovery(provider) {
+	zonesDiscoveryEnabled := s.providerSpec.ZonesDiscovery(provider)
+
+	azureSuffixNeeded := provider == runtime.Azure && s.useMachineImageVersionSuffix
+	if !zonesDiscoveryEnabled && !azureSuffixNeeded {
```
jaroslaw-pieszka, line 51: The message below is misleading given the condition.

internal/process/steps/discover_available_zones_cb.go
```diff
@@ -43,12 +45,17 @@ func (s *DiscoverAvailableZonesCBStep) Name() string {
...
 		log.Info(fmt.Sprintf("Zones discovery disabled for provider %s, skipping", provider))
 		return operation, 0, nil
 	}
-	if len(operation.DiscoveredZones) > 0 {
-		log.Info("Available zones already discovered, skipping")
+	zonesAlreadyDone := !zonesDiscoveryEnabled || len(operation.DiscoveredZones) > 0
+	suffixesAlreadyDone := !azureSuffixNeeded || len(operation.MachineImageVersionSuffixes) > 0
+	if zonesAlreadyDone && suffixesAlreadyDone {
+		log.Info("Zones and machine image version suffixes already discovered, skipping")
```
jaroslaw-pieszka, line 58: I have doubts since if we have discovery disabled, suffixes not needed we enter this path as well.

internal/process/steps/discover_available_zones_cb.go
```diff
@@ -87,33 +94,36 @@ func (s *DiscoverAvailableZonesCBStep) Run(operation internal.Operation, log *sl
...
-		zones, err := client.AvailableZones(context.Background(), machineType)
-		if err != nil {
-			return s.operationManager.RetryOperation(operation, fmt.Sprintf("unable to get available zones for machine type %s", machineType), err, 10*time.Second, time.Minute, log)
+	for machineType := range machineTypes {
+		if zonesDiscoveryEnabled {
+			zones, err := client.AvailableZones(context.Background(), machineType)
+			if err != nil {
+				return s.operationManager.RetryOperation(operation, fmt.Sprintf("unable to get available zones for machine type %s", machineType), err, 10*time.Second, time.Minute, log)
+			}
+			rand.Shuffle(len(zones), func(i, j int) { zones[i], zones[j] = zones[j], zones[i] })
```
jaroslaw-pieszka: This is quite cryptic so I would extract this part (it is duplicated somewhere in providers as well), giving some meaningful name to the function.

internal/process/steps/discover_available_zones_cb.go
KsaweryZietara: I'll update the message to reflect both conditions.

internal/process/steps/discover_available_zones_cb.go
KsaweryZietara: You're right that `zonesAlreadyDone && suffixesAlreadyDone` is technically `true` when both are disabled/not needed, but that case is already handled by the first guard (`!zonesDiscoveryEnabled && !azureSuffixNeeded`), so we never reach the second block in that state.

internal/process/steps/discover_available_zones_cb.go
KsaweryZietara: I extracted this into a helper function and added a comment explaining its purpose:

```go
// shuffleZones randomizes zone order so that different instances spread across
// availability zones instead of all landing in the first zone returned by the API.
func shuffleZones(zones []string) {
	rand.Shuffle(len(zones), func(i, j int) { zones[i], zones[j] = zones[j], zones[i] })
}
```

jaroslaw-pieszka (approved):

PR #3469, opened 2026-07-30
Remove trial limit per GA

marco-porru:
**Description**

- remove limit of 1 instance per GA

**Related issue(s)**


PK85 (changes requested):
This is already supported in the configuration. Rejected

